PR #3, opened 2026-04-14
fix(dashboard): show meaningful tail of project name in Cost by Project

shudonglin:
## Summary
The **Cost by Project** chart on the dashboard was left-truncating project names, dropping the most identifying part. For example, `claude:HubLogic-HanHan-sub2apipay` rendered as `ic-HanHan-sub2apipay`, and `claude:HubLogic-Papers` as `aude:HubLogic-Papers`.

The cause is at `Dashboard.tsx:418`:
```ts
const label = c.project.length > 20 ? c.project.slice(-20) : c.project;
```
`slice(-20)` keeps the **last** 20 characters. The `BarRow` label column already has `overflow: hidden; text-overflow: ellipsis` in a 180px-wide flex item with `title=` for hover, so the manual slice was both unnecessary and actively harmful.

### Fix
Replace with a path-aware shortener (`displayProject`):
1. Strip the redundant `<source>:` prefix (`claude:`, `codex:`, …) — the chart is per-project, so the source is implied.
2. Keep adding trailing dash-segments from the right until the next one would overflow the column (~22 chars). Path separators and dashes-inside-folder-names look identical in the project id, so we err on the side of context: `HubLogic-Codes-hublogic-landing-page` → `hublogic-landing-page`, not just `page`.
3. Keep the full project id in the row's `title` tooltip for disambiguation when hovering.

`BarRow` gains an optional `title?: string` prop so callers can render a different value on hover than what's shown. Defaults to the label, so all existing call sites are unchanged.

### Before / After

| Project id | Before (broken) | After |
|---|---|---|
| `claude:HubLogic-HanHan-sub2apipay` | `ic-HanHan-sub2apipay` | `HanHan-sub2apipay` |
| `claude:HubLogic-Codes-hublogic-landing-page` | `hublogic-landing-page` * | `hublogic-landing-page` |
| `claude:HubLogic-Papers` | `aude:HubLogic-Papers` | `HubLogic-Papers` |
| `claude:MRI Health Analysis` | `:MRI Health Analysis` | `MRI Health Analysis` |

\* the long names were also being CSS-clipped from the right after the slice, producing things like `HubLogic-Codes-hublogic-la…`

Full project id (with `claude:` / `codex:` prefix and parent path) is always available on hover.

## Test plan
- [x] `npm run build` — TypeScript + vite build green
- [x] Verified live in browser at localhost:8384/dashboard:
  - All 8 rows render the meaningful tail-segments
  - No row needs CSS ellipsis clipping with the new shortener (`scrollWidth <= clientWidth` for every row in the test data)
  - Hover tooltip shows the full `claude:...` / `codex:...` id

### DIFF
--- a/clawjournal/web/frontend/src/views/Dashboard.tsx
+++ b/clawjournal/web/frontend/src/views/Dashboard.tsx
@@ -42,8 +42,9 @@ function shortModel(model: string): string {
   return raw;
 }
 
-function BarRow({ label, value, max, total, color = colors.blue400, fmt }: {
+function BarRow({ label, title, value, max, total, color = colors.blue400, fmt }: {
   label: string;
+  title?: string;
   value: number;
   max: number;
   total?: number;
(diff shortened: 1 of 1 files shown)


shudonglin:
The names of "Cost by Project" dashboard are truncated, and not meaningful. Changed it to more meaningful display with hover for full names like the "Focus Map". @kai-rayward 

Before
cost-by-project

After
cost-by-project-after

PR #13, opened 2026-04-21
fix(share): align token totals to input+output, matching SessionDetail

shudonglin:
## Summary

- The Share queue displayed only `input_tokens` (`Share.tsx:1427,1556,1832,2132` and the two `reduce` accumulators at `:959,:1234`). Prompt caching drops raw `input_tokens` to near zero on long sessions — one real trace of 207 messages / 200k output tokens / 28M cache-read tokens rendered as **0k tokens**.
- Fix aligns the six Share-tab display and aggregation sites to `input_tokens + output_tokens`, matching the formula already used in `SessionDetail.tsx:294` and `clawjournal inbox --json`.
- `get_share_ready_stats` (`workbench/index.py`) and the `shareReady()` API return type now include `output_tokens` so the frontend can sum without a second round-trip.

## Test plan

- [x] `pytest tests/workbench/` — all 259 pass locally
- [x] `npm run build` inside `clawjournal/web/frontend` — compiles clean (`tsc -b && vite build`)
- [ ] Open `/share` on a workbench with ≥1 heavily-cached session; the draft-bundle row tokens and the header total are no longer `0k` / `2k` but reflect input + output
- [x] `SessionDetail` "Tokens" row for the same session matches what `/share` now shows

### DIFF
--- a/clawjournal/web/frontend/src/api.ts
+++ b/clawjournal/web/frontend/src/api.ts
@@ -160,7 +160,7 @@ export const api = {
     return request('/projects');
   },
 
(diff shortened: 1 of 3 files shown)


shudonglin:
for this specific session, "the readme file is a bit confusing by mixing the python command and skills..."
it shows 200.8k tokens used
200k_token
but in the /share page, it shows 0k token.
before_0_token

raised this pr to fix this bug to align the token calculation to include both input and output token in all the caculation. @kai-rayward 
after_201k_token

PR #14, opened 2026-04-21
feat(share): Review-step ergonomics — bulk-include, remove-undo, restore drawer

shudonglin:
## Summary

Two UX friction points in the Share wizard's Review step, addressed in a single client-side change to `Share.tsx` and a backward-compatible extension to `Toast.tsx`.

**Problem 1 — Bulk-include was nearly unusable.** `Include all clean (N)` requires `classify() === 'clear'`, i.e. zero AI-flagged items *and* zero residual redactions. In real traces nearly every session carries at least one AI flag (org/person names), so the button sat at `(0)` most of the time and users had to approve each card manually.

**Problem 2 — Remove had no safety net.** `Remove from bundle` called `removeFromQueue` which filtered the id out of `queueOrder` immediately. The session vanished from view. The only recovery was going back to Queue and re-picking from "Add more traces" (potentially 100+ candidates). Misclicks were unrecoverable.

### 1. Tiered bulk-include

A second button next to the existing "Include all clean": **"Include all with flags (N) ⚠"**. Eligibility: `!approvedIds.has(id) && classify() !== 'clear'`. A hover tooltip breaks down the flag composition ("N AI-flagged (names, orgs, etc.)", "N with residual redactions"). The existing `Include all clean` button keeps its exact behavior and eligibility — no regression.

Both bulk actions now fire an **8s undo toast** ("Included N traces · Undo") that reverses only the ids *that bulk added* (manual approvals stay).

### 2. Remove safety net

Two layers, shared between Queue step and Review step:

- **Transient undo toast** (8s): every `Remove from bundle` click shows "Removed '…' · Undo" bottom-right. Click Undo → session restored to its exact previous slot via a captured `originalIndex` (clamped).
- **Persistent removed-drawer**: below the card list (on both Queue and Review steps). Collapsed by default with count (`▸ N removed from bundle`). Expanded shows per-entry title + relative time + individual `Restore` button. A **`Restore all`** button on the drawer header puts every removed entry back in one click (sorted by `originalIndex` for deterministic placement).

### 3. Cache preservation across remove/restore

The existing useEffect that purged `redactedSessions[id]` and `approvedIds` when an id left `queueOrder` was too aggressive for the new flow. Restored a removed session → expensive AI redaction re-ran and the user's approval status was lost. Fix keeps entries whose id is in the queue **OR** in the removed-drawer. Cache is purged only when the session genuinely leaves the wizard (neither queued nor removed).

### 4. Toast component

Extended `toast(message, typeOrOptions?)` to accept `{ type, duration, action: { label, onClick } }` as the second arg. All existing callers (`toast('msg')`, `toast('msg', 'info')`, `toast('msg', 'success')`) are unaffected — the union preserves the string overload.

## Scope

- **Client-side only.** No backend, no API, no DB, no schema.
- Only two files touched: `clawjournal/web/frontend/src/components/Toast.tsx` (+47 / -5), `clawjournal/web/frontend/src/views/Share.tsx` (+311 / -9).
- No new tests — the frontend has no test framework yet; verified live against a populated workbench (163 indexed sessions).

## Test plan

- [ ] `cd clawjournal/web/frontend && npm run build` passes cleanly.
- [ ] `/share` Queue step: click any Remove → toast with Undo → click Undo within 8s → session returns to its original slot.
- [ ] After 8s timeout: session stays in drawer (`▸ 1 removed from bundle`). Expand, click Restore → session returns.
- [ ] Click `Include all with flags (N) ⚠` on Review step → all flagged traces become approved; toast shows `Included N flagged traces · Undo`; clicking Undo removes only those ids from approvedIds.
- [ ] Hover the `⚠` button → tooltip lists AI-flagged vs residual counts.
- [ ] Remove all sessions from the queue → empty-bundle state appears on Queue step → drawer is still visible there with `Restore all` → click it → all sessions restored at their original indices.
- [ ] Approve + remove a session → restore → approval + redaction cache preserved (no re-run).
- [ ] Successful package clears the drawer.

## Screenshots

*(intentionally omitted — happy to add if useful; the change is navigable via the live workbench)*

### DIFF
--- a/clawjournal/web/frontend/src/components/Toast.tsx
+++ b/clawjournal/web/frontend/src/components/Toast.tsx
@@ -1,14 +1,26 @@
 import { createContext, useContext, useState, useCallback, useRef } from 'react';
 import { colors } from '../theme.ts';
 
+export interface ToastAction {
+  label: string;
+  onClick: () => void;
+}
+
+export interface ToastOptions {
+  type?: 'success' | 'error' | 'info';
+  duration?: number;          // ms, default 3500
+  action?: ToastAction;
+}
+
 interface ToastItem {
   id: number;
   message: string;
   type: 'success' | 'error' | 'info';
+  action?: ToastAction;
 }
 
 interface ToastCtx {
-  toast: (message: string, type?: 'success' | 'error' | 'info') => void;
+  toast: (message: string, typeOrOptions?: 'success' | 'error' | 'info' | ToastOptions) => void;
 }
 
 const Ctx = createContext<ToastCtx>({ toast: () => {} });
@@ -21,10 +33,18 @@ export function ToastProvider({ children }: { children: React.ReactNode }) {
   const [items, setItems] = useState<ToastItem[]>([]);
   const counter = useRef(0);
 
-  const toast = useCallback((message: string, type: 'success' | 'error' | 'info' = 'info') => {
+  const toast = useCallback((
+    message: string,
+    typeOrOptions?: 'success' | 'error' | 'info' | ToastOptions,
+  ) => {
+    const opts: ToastOptions =
+      typeof typeOrOptions === 'string' ? { type: typeOrOptions }
+      : typeOrOptions ?? {};
     const id = ++counter.current;
-    setItems(prev => [...prev, { id, message, type }]);
-    setTimeout(() => setItems(prev => prev.filter(t => t.id !== id)), 3500);
+    const type = opts.type ?? 'info';
+    const duration = opts.duration ?? 3500;
+    setItems(prev => [...prev, { id, message, type, action: opts.action }]);
+    setTimeout(() => setItems(prev => prev.filter(t => t.id !== id)), duration);
   }, []);
 
   const bgMap = { success: colors.green700, error: colors.red500, info: colors.gray800 };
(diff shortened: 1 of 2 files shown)


shudonglin:
the original include all rule seems to be too tight, I added a include all with flag.
and also added a restore session after remove it, to prevent accidentally removal. @kai-rayward 

before
before_need_include_one_by_one

after
remove_restore_bundles

shudonglin:
the review page had changed completely. This PR is not relevant any more. I am dropping it.  FYI @kai-rayward 

image

shudonglin:
Update: rebased onto latest `main` (picks up #13, #15, #16, #17, #18, #20) and added a follow-up fix commit addressing two bugs surfaced by a Codex code review:

**`02bcaa1` — fix(share): preserve restore order and exclude in-flight retries from bulk-approve**

1. **Multi-item undo corrupted queue order.** The original `removeFromQueue` captured `originalIndex` from the already-mutated `queueOrder`, so restoring more than one removed trace reshuffled the bundle. Example: removing `B` then `D` from `[A,B,C,D]` stored indices `1` and `2`; `Restore all` then rebuilt `[A,B,D,C]`. Switched to a neighbor-anchor model — record `beforeId`/`afterId` at removal time, restore newest-first, prefer `afterId` when both are present. Verified across 17 deterministic scenarios (adjacent, non-adjacent, reverse-order, head, tail, remove-all).

2. **Bulk "Include all with flags" could approve in-flight retries.** Filter used `classify(...) !== 'clear'`, which matches both `review` and `checking`. Clicking `Retry AI` puts a trace back into `checking`; if the user then clicked `Include all with flags` before the retry returned, it was approved without showing the refreshed redaction. Tightened to `=== 'review'`.

Diff scope unchanged from the original PR (only `clawjournal/web/frontend/src/views/Share.tsx`). Type-check and lint clean.

PR #43, opened 2026-05-27
fix: rebalance search and share layouts

shudonglin:
## Summary
- rebalance the `/search` shell so the input and empty state read correctly on desktop
- widen the `/share` flow and keep the outer shell stable across steps while preserving narrow package/done cards
- center the empty queue states without stretching the CTA controls

## Verification
- `npm run build` in `clawjournal/web/frontend`
- manually checked `http://127.0.0.1:8384/search`
- manually checked `http://127.0.0.1:8384/share`

## Claude Review
- initial Claude diff review flagged two real layout issues: the brittle search empty-state height and the share step-shell width jump
- both issues were fixed before opening/updating this PR
- final Claude pass left one non-blocking note: the queue empty-state cards use centered flex-column layout, which could be fragile if future children need full-width block layout
- manually verified the current queue empty-state content in-browser; the current icon, copy, and CTA render correctly, so this is not a blocking issue for the present UI

## Notes
- no new tests added for this change set

### DIFF
--- a/clawjournal/web/frontend/src/views/Search.tsx
+++ b/clawjournal/web/frontend/src/views/Search.tsx
@@ -7,6 +7,8 @@ import { EmptyState } from '../components/Spinner.tsx';
 import { useToast } from '../components/Toast.tsx';
 import { colors } from '../theme.ts';
 
+const SEARCH_SHELL_WIDTH = 1120;
+
 export function Search() {
   const { toast } = useToast();
   const [query, setQuery] = useState('');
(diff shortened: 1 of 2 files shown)


shudonglin:
Claude review summary for this PR:

- initial pass flagged two real issues: a brittle search empty-state height and a share flow shell-width jump between steps
- both were fixed on this branch before finalizing the PR
- final pass left one non-blocking note: the queue empty-state cards use centered flex-column layout, which could be fragile if future children expect full-width block layout
- manually verified the current queue empty-state content in-browser; current icon/copy/CTA render correctly, so I am treating that note as non-blocking

shudonglin:
just a cosmetic pr for the html report formatting.

PR #58, opened 2026-05-31
fix(redaction): stop TruffleHog gate blocking on generic-Azure false positives

shudonglin:
## Summary

The mandatory TruffleHog gate blocks **every** share export that contains terminal output, because TruffleHog's **legacy generic `Azure` detector (type 3)** fires on ordinary agent-session content and always comes back `unverified`. A user hit:

```
TruffleHog blocked the share: 3 finding(s) (verified=0, unverified=3, unknown=0).
Examples: L3 unverified Azure 127.***432:, L3 unverified Azure scri***\n??, L5 unverified Azure \u00***0m\n
```

Reproduced against the real blocked bundle — the three "Azure" matches were:
- `127.0.0.1:51678->127.0.0.1:5432:` (PLAIN) — a localhost Postgres connection line
- `scripts/seed-routing-rules.sh\n??` (PLAIN) — a file path
- `[90mnull[0m[0m\n` (BASE64) — ANSI color codes

All three are structural noise, never real secrets — and since the gate blocks on *any* finding, traces with terminal output (i.e. almost all of them) can't be shared.

## Fix

Suppress **only the `unverified`** findings from the generic `Azure` detector in `scan_file`'s parse loop (new `NOISY_UNVERIFIED_DETECTORS`), keeping its `verified`/`unknown` findings and **every specific Azure detector** (`AzureStorage`, `AzureSAS`, …) fully active.

This is deliberately narrower than the existing `EXCLUDED_DETECTORS` (which disables a detector entirely, and per its own docstring is reserved for detectors that "never catch real leaks", e.g. `refiner`). Azure *can* catch real secrets and `secrets.py` has no Azure fallback, so a blanket exclusion would let **verified** Azure secrets through. Verified `AzureStorage` connection strings are still caught alongside this change.

## Tests

- `test_unverified_generic_azure_false_positives_are_suppressed` (RED→GREEN, reproduces the exact `127.***432:` finding)
- `test_verified_generic_azure_still_blocks` — suppression is surgical
- `test_specific_azure_detector_unverified_still_blocks` — only the generic detector is treated as noisy

Full suite green (1826). The user's real blocked bundle now scans `blocking=False, findings=0`.

### DIFF
--- a/clawjournal/redaction/trufflehog.py
+++ b/clawjournal/redaction/trufflehog.py
@@ -73,6 +73,21 @@ def _scrubbed_subprocess_env() -> dict[str, str]:
 #   returns ``unverified`` for those, so they are never real leaks.
 EXCLUDED_DETECTORS: tuple[str, ...] = ("refiner",)
 
+# Detectors whose UNVERIFIED findings are dropped as structural false
+# positives, while their verified/unknown findings still block. This is
+# narrower than EXCLUDED_DETECTORS (which disables a detector entirely,
+# reserved for detectors that never catch real leaks): the detector stays
+# active, so a real, live secret it confirms still blocks the share.
+#
+# - ``Azure`` is TruffleHog's legacy generic Azure detector (type 3). Its
+#   broad pattern, fed through the PLAIN and BASE64 decoders, matches ordinary
+#   agent-session terminal content — ``127.0.0.1:5432`` connection lines, file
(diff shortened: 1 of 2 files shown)


shudonglin:
Folded into #59 as a single combined commit (Azure-gate suppression + ANSI/terminal-noise stripping). Closing in favor of #59.

PR #59, opened 2026-05-31
fix(share): unblock clean shares (TruffleHog) + download-filename, submit-button & routing fixes

shudonglin:
## Summary

Several issues blocked or confused the share/export path. The branch now includes the original share-flow fixes plus a forward fix that keeps the TruffleHog export gate fail-closed.

## 1. TruffleHog gate — reduce clean-share terminal noise without broad detector suppression
`redaction/trufflehog.py`, `redaction/normalize.py`, `findings.py`, `workbench/index.py`

The mandatory post-redaction gate still blocks on **any** TruffleHog finding, including unverified generic `Azure` findings. Earlier in this branch we considered dropping all `Azure` + `unverified` findings, but that was too broad for the export privacy boundary.

- Strip terminal/ANSI control sequences before scanning (`strip_terminal_control_sequences`) so rendering artifacts do not bloat bundles or create avoidable detector noise.
- Walk the **widened** message fields (author/invocations/snippets/extra) so the gate, the engine, and finding preview see the same payload.
- Keep generic Azure findings fail-closed: verified, unknown, and unverified TruffleHog hits all remain visible/blocking unless a detector is explicitly excluded.

## 2. Download zip → bare-UUID filename
`web/frontend/src/api.ts`

"Download zip" saved the bundle as a bare UUID with no `.zip` extension (bytes were correct; only the suggested filename was lost). Cause: `URL.revokeObjectURL(url)` ran synchronously right after `a.click()`, and Chromium reads the blob asynchronously, dropping the filename.

- Defer the revoke (60s), add `rel="noopener"`, parse RFC 5987 `filename*` (the general `charset'lang'value` form) **safely** (try/catch), and fall back to a `.zip`-bearing name.

## 3. Submit button flicker + premature Done routing
`web/frontend/src/views/Share/DoneStep.tsx`, `index.tsx`

The "Submit to ClawJournal Research" button toggled present/absent for the **same URL** because `shareDestination` (loaded by an async, network-dependent probe) collapsed loading / failed / not-configured into `null`.

- Track `destinationLoading` / `destinationFailed`, load the probe independently, and render a spinner / Retry / button accordingly.
- Because the probe is now independent, the package→done and submit-deeplink routing effects could decide submit-vs-done before it resolved and route to Done irreversibly. Both are gated on `!destinationLoading`.

## Review & verification

Verified the PR head merged cleanly into latest `origin/main` (`accab5d`). On that merge result:

- `PYTHONPATH=. /tmp/clawjournal-pr59-review-venv/bin/python -m pytest -q`: **1931 passed**.
- `npm ci && npm run build`: passes.
- `git diff --cached --check`: passes.

Also ran targeted TruffleHog/normalization/share-gate tests on the PR branch after the forward fix.

### DIFF
--- a/clawjournal/findings.py
+++ b/clawjournal/findings.py
@@ -523,23 +523,32 @@ def _resolve_field_text(blob: dict[str, Any], finding: Finding) -> str | None:
 
     # tool_uses[<idx>].<input|output> or tool_uses[<idx>].<input|output>.<key>
     match = re.match(r"tool_uses\[(\d+)\]\.(\w+)(?:\.(.+))?$", field)
-    if not match:
-        return None
-    tool_idx = int(match.group(1))
-    branch = match.group(2)
-    nested_key = match.group(3)
-    tool_uses = msg.get("tool_uses")
-    if not isinstance(tool_uses, list) or tool_idx >= len(tool_uses):
-        return None
-    tool = tool_uses[tool_idx]
-    if not isinstance(tool, dict):
-        return None
-    value = tool.get(branch)
-    if nested_key:
-        if isinstance(value, dict) and isinstance(value.get(nested_key), str):
-            return value[nested_key]
-        return None
-    return value if isinstance(value, str) else None
+    if match:
+        tool_idx = int(match.group(1))
+        branch = match.group(2)
+        nested_key = match.group(3)
+        tool_uses = msg.get("tool_uses")
+        if not isinstance(tool_uses, list) or tool_idx >= len(tool_uses):
+            return None
+        tool = tool_uses[tool_idx]
+        if not isinstance(tool, dict):
+            return None
+        value = tool.get(branch)
+        if nested_key:
+            if isinstance(value, dict) and isinstance(value.get(nested_key), str):
+                return value[nested_key]
+            return None
+        return value if isinstance(value, str) else None
+
+    # Widened message model (author / invocations[...] / snippets[...] /
+    # extra.*). The labels come from iter_widened_text_locations; resolve by
+    # re-walking and matching the label rather than re-parsing the path.
+    from .parsing.widened import iter_widened_text_locations  # noqa: PLC0415 — lazy
+
+    for text, label in iter_widened_text_locations(msg):
+        if label == field:
+            return text
+    return None
 
 
 def derive_preview(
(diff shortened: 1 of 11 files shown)


shudonglin:
### Added two share-flow UI fixes to this branch

While testing the share flow end-to-end, two more bugs surfaced and are fixed here:

- **`c10d3b4` — Download zip saved a bare-UUID file with no `.zip` extension.** The bytes were correct, but `download()` revoked the blob object URL synchronously right after `click()`; Chromium reads the blob asynchronously, so the suggested filename was dropped. Fixed by deferring the revoke, adding `rel="noopener"`, an RFC 5987 `filename*` parser, and a `.zip`-bearing fallback name.
- **`f31b466` — "Submit to ClawJournal Research" button appeared/disappeared for the same URL.** It's gated on `shareDestination`, loaded by an async, network-dependent probe. Collapsing "loading", "transient failure", and "not configured" into `null` rendered a buttonless final page (then the button popped in late, or never, after a daemon restart). Fixed by tracking `destinationLoading`/`destinationFailed` and loading the probe independently — the Done step now shows a spinner while loading, a Retry on failure, and the button once resolved.

Both are frontend-only (`tsc` build passes; Python suite unaffected).

shudonglin:
## Codex review — GATE: PASS ✅ (after one fix round)

Independent review via `codex --yolo -p newapi` (model `gpt-5.5`, `xhigh` reasoning) on the two share-flow UI fixes added to this branch.

**First pass: GATE FAIL** — surfaced two real issues, both fixed:

| Sev | Finding | Fix |
|-----|---------|-----|
| P2 | Making the hosted-destination probe independent meant the package→done and submit-deeplink routing effects decided submit-vs-done while the probe was still loading (`shareDestination` null), routing to **Done** and never recovering once upload became available. | Gate both effects on `!destinationLoading` (and add it to deps) so routing waits for the probe. |
| P3 | `decodeURIComponent(filename*)` could throw on malformed percent-encoding and abort an otherwise valid download; only handled the `UTF-8''` form. | Use the general `charset'lang'value` regex and wrap decode in try/catch with a plain/`.zip` fallback. |

**Second pass: GATE PASS** — P1/P2/P3 all none. Codex verified the package routing reruns when `destinationLoading` flips false (no stuck-on-Package), the submit deep-link fallback waits with correct deps, the filename parse no longer throws, and `npm run build` passes.

All commits on this branch squashed into one (`c5f9552`); `pytest` on the touched areas: 111 passed.

kai-rayward:
@shudonglin I updated as the TruffleHog change above may be too much, so make it more conservative.

PR #60, opened 2026-05-31
fix(parsing): standardize project names to cwd basename across agents

shudonglin:
## Summary

Standardize the `project` field to the **directory basename** for all agents. Claude (and Cursor, which reuses Claude's builder) labeled sessions with the full path below the home dir, while Codex/OpenCode/OpenClaw/Kimi use just the basename — so the same physical project showed up under two different names depending on which agent ran in it.

| Agent | Before | After |
|---|---|---|
| Codex / OpenCode / OpenClaw / Kimi | `codex:llm-gateway-infra` | `codex:llm-gateway-infra` (unchanged) |
| **Claude** | `claude:Rayward-Codes-llm-gateway-infra` | **`claude:llm-gateway-infra`** |

## Why it wasn't a one-liner

Claude derives the project from its hyphen-encoded session dir name (`-Users-me-Rayward-Codes-llm-gateway-infra`). That encoding is **ambiguous** when a folder name itself contains hyphens (`llm-gateway-infra` is indistinguishable from `llm/gateway/infra`), so the basename can't be recovered from the dir name alone.

## Fix

Read the real absolute `cwd` recorded in the Claude session content and use `Path(cwd).name` — falling back to the old dir-name heuristic when no `cwd` is present. New `_read_project_cwd()` helper.

## Impact / migration
- **Backfill is automatic:** the scanner full-re-parses every project each pass (`INSERT OR REPLACE`), so a re-scan updates existing sessions' `project` values.
- **Config caveat:** confirmed/excluded-project config referencing old full-path Claude names won't auto-migrate (the same cwd-ambiguity means they can't be mapped reliably) and may need re-confirming.
- **Cursor limitation:** Cursor shares the ambiguous dir-name encoding but its session content carries **no `cwd`**, so deep Cursor paths can't be reduced to a basename and stay on the heuristic.

## Tests
- `_read_project_cwd` unit tests (reads cwd, None when absent/empty).
- Regression: a hyphenated project folder (`llm-gateway-infra`) now yields `claude:llm-gateway-infra`, not the leaked `claude:Rayward-Codes-...`.
- Updated two existing fixtures whose `cwd` didn't match their dir name. Full suite green (1902).

Branched off `main`, independent of the TruffleHog/share PR.

### DIFF
--- a/clawjournal/parsing/parser.py
+++ b/clawjournal/parsing/parser.py
@@ -831,6 +831,29 @@ def parse_project_sessions(
     elif not locator:
         native_dir = PROJECTS_DIR / project_dir_name
 
+    # Prefer the basename of the real cwd recorded in the session content. The
+    # hyphen-encoded dir name can't be split back into a basename when a folder
+    # name itself contains hyphens (e.g. "llm-gateway-infra"), which is why the
+    # dir-name heuristic leaks the full path (claude:Rayward-Codes-...). This
+    # keeps Claude consistent with codex/opencode/openclaw (basename only).
+    # Cover every Claude session location: native root sessions, subagent-only
+    # transcripts (read recursively below native_dir), and local-agent nested
+    # project dirs — all of which record an absolute cwd.
+    if not project_dir_name.startswith("_cowork_"):
+        cwd_dirs: list[Path] = []
+        if native_dir:
+            cwd_dirs.append(native_dir)
+        if locator:
+            for la_session in locator.get("local_agent_sessions", []):
+                nested = la_session.get("nested_project_dir")
+                if nested:
+                    cwd_dirs.append(nested)
+        cwd = _read_project_cwd(*cwd_dirs)
+        if cwd:
+            # `or cwd` guards a root cwd ("/") whose basename is empty, matching
+            # the codex/opencode builders.
+            project_name = f"claude:{Path(cwd).name or cwd}"
+
     if native_dir and native_dir.exists():
         for session_file in sorted(native_dir.glob("*.jsonl")):
             parsed = _parse_claude_session_file(session_file, anonymizer, include_thinking)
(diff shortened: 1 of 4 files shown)


shudonglin:
## Codex review — GATE: PASS ✅

Independent review via `codex` (model `gpt-5.5`, `xhigh` reasoning) on the project-basename fix.

**Verdict:** GATE PASS — **P1: none, P2: none.**

Verified end-to-end:
- Native root, **subagent-only**, and **local-agent-only** Claude session labels now recover the cwd basename (`claude:llm-gateway-infra`) instead of leaking the hyphen-encoded full path (`claude:Rayward-Codes-llm-gateway-infra`).
- Root-cwd empty-basename guard (`Path(cwd).name or cwd`) is correct.
- Cursor docstring issue is gone (Cursor intentionally stays on the dir-name heuristic — its session files carry no cwd).
- `pytest tests/parsing/test_parser.py` → **157 passed**; full suite **1907 passed**.

### Findings addressed since the prior round
| Sev | Finding | Resolution |
|-----|---------|------------|
| P1 | cwd override only fired when `native_dir` existed → subagent-only / local-agent-only projects skipped | `_read_project_cwd` now takes multiple candidate dirs and searches **root sessions first, then recursive `rglob`**; call site feeds it `native_dir` + every local-agent `nested_project_dir` |
| P2 | empty-basename: `Path(cwd).name` collapses `/` → `claude:` | guard `Path(cwd).name or cwd` |
| P2 | docstring misleadingly claimed Cursor coverage | rewritten; commit message documents the Cursor limitation |
| P3 | nested `rglob()` is lazy — `OSError` during iteration uncaught + unordered | materialize `sorted(project_dir.rglob("*.jsonl"))` inside the `try` |

### Remaining P3 (acknowledged, not actioned)
- LA-only test hand-builds the `locator`, so it exercises `parse_project_sessions` but not the full `discover_projects` plumbing for the hyphenated case. The recovery logic itself is covered at the `parse_project_sessions` level; `discover_projects.display_name` is a separate, pre-existing surface (used for `config --exclude` matching) and out of scope for this fix.

Squashed to a single commit `0f94afa`.

shudonglin:
Screenshot 2026-05-31 at 9 44 52 PM
This screenshot is before the basename fix.

align the directory basename for all sessions, including codex, claude code, openclaw etc.

kai-rayward:
@shudonglin pushed some fixes (generated by codex), and feel free to merge it.

PR #80, opened 2026-06-12
Add clawjournal trufflehog install (managed, pinned, checksum-verified share-gate binary)

kai-rayward:
## Summary

TruffleHog install is the one genuine extra setup step for students — the share gate fails closed without it. This adds a managed install path so nobody has to find brew or the upstream release page:

- **`clawjournal trufflehog install [--force]`** — downloads the pinned **v3.95.5** from trufflesecurity's official GitHub release for macOS / Linux / Windows (x86-64 and ARM64), verifies the archive against **sha256 checksums vendored in the source** (never trusts what the network served), and installs into `~/.clawjournal/bin/` with a **stage → fsync → verify → atomic publish** flow: the binary is extracted to a temp file, sanity-checked with `--version`, and only then `os.replace()`d into place — a bad download can never displace a working managed binary. No root needed. Human mode streams download progress (size + a tick every 16 MB).
- **`clawjournal trufflehog status [--json]`** — shows which binary the share gate will use (bare `version` matching `install --json`, plus the raw engine `fingerprint`). Exit code 1 when no binary resolves, consistently across human and JSON modes.
- **`resolve_binary()`** in `redaction/trufflehog.py` prefers the managed copy over `PATH` (pinned + verified beats whatever brew last installed). The share gate, findings engine, and `engine_fingerprint` all resolve through it; the fingerprint cache keys on path+mtime+size, so a long-running daemon picks up the install without a restart.

### Correctness & hardening

- **`already-installed` means "the gate will use it AND it is the pin"**: the early exit requires `is_file` + executable + a `--version` match against `PINNED_VERSION`. Non-executable strays, directories, broken copies, and off-pin binaries from older clawjournals are repaired by a fresh install instead of reporting success while the gate stays blocked.
- **`install()` never raises**: every failure returns a status dict and fails closed — truncated chunked downloads (`IncompleteRead`), truncated gzip streams, a stray file blocking the bin dir, oversized payloads (200 MB download / 500 MB decompression caps).
- **Failure domains are reported distinctly**: archive parsing errors → `archive-invalid` (re-download might help); local file I/O and publish errors (disk full, directory squatting on the target, Windows refusing to replace a running `.exe`) → `install-failed` without the misleading GitHub URL. `download-failed` / `checksum-mismatch` results carry actionable `hint` text (TLS-intercepting campus/corporate proxies, `selfupdate`, manual fallback); the CLI prints error + URL + hint.
- **Exact-name tar member extraction** (no path traversal possible); the Windows archives were verified to contain a member named exactly `trufflehog.exe`, and every vendored checksum (including `windows_arm64`) was re-verified against upstream's release `checksums.txt`.
- **Windows-safe on Python 3.10+**: `os.chmod` on the temp path instead of `os.fchmod` (absent on Windows before 3.13).
- **AGPL posture unchanged**: fetched from upstream's own release artifacts at the user's explicit request, invoked subprocess-only with the existing scrubbed env.

### Remediation texts updated

`INSTALL_HINT`, events doctor, `install.sh` / `install.ps1` (which now also detect the managed copy instead of probing PATH only, and print the venv-qualified command since the venv isn't on `PATH` at that point), README, and PRIVACY.md all lead with the new command.

### Review process

Two multi-agent adversarial review rounds (findings only accepted on majority verdict from independent judges):
- **Round 1** (67 agents, during development): 10 confirmed findings fixed — `os.fchmod`-on-Windows crash, never-raises contract holes, stale brew-only remediation texts, missing test coverage.
- **Round 2** (56 agents, fresh lenses on the committed PR): 9 confirmed findings fixed in ca9a534 — the already-installed/X_OK/pin hole, verify-after-publish destroying good binaries, failure-domain misclassification, `status --json` exit-code asymmetry, missing `windows_arm64`, silent downloads, hint-less errors, version-shape inconsistency, bare-command script hints. Two findings were judged real-but-fine-as-is and deliberately left (benign double-resolution race in `scan_file`; bin-dir umask).

## Test plan

- 50 offline installer tests (fake urlopen + in-memory tarballs): happy path, checksum mismatch fails closed, pin-aware already-installed, reinstall-on-broken/off-pin, verify-failure-preserves-existing-binary, replace-failure classification, unsupported platform, download failures incl. truncation, size caps, missing/traversal/corrupt archive members, never-raises backstop, the real `_verify_installed_binary` contract, progress callback, hint plumbing, resolution precedence, CLI dispatch + exit codes + argparse round-trip, and an 8-case OS/arch matrix.
- Full suite: **2225 passed**; CI green on 3.10–3.13 + wheel smoke for both commits.
- Verified live end-to-end on macOS arm64: real download → checksum verification → install → status flip from brew 3.94.3 (PATH) to managed 3.95.5 → real `scan_file` through the managed binary → pin-aware `already-installed` re-running the real binary.

### DIFF
--- a/.github/workflows/trufflehog-pin.yml
+++ b/.github/workflows/trufflehog-pin.yml
@@ -1,0 +1,65 @@
+name: TruffleHog Pin Integrity
+
+# Verifies that the sha256 checksums vendored in
+# clawjournal/redaction/trufflehog_install.py still match the upstream
+# release they pin. Unit tests are hermetic and cannot see this — only a
+# real fetch of upstream's checksums.txt catches a bad pin bump or a
+# release asset that was silently re-uploaded.
+#
+# IMPORTANT: do NOT add this workflow's check to branch-protection
+# required checks. It is path-filtered at the trigger level, so on PRs
+# that do not touch the pinned files it produces no check run at all — a
+# required status would then sit "Expected" forever and block the PR.
+
+on:
+  pull_request:
+    paths:
+      - 'clawjournal/redaction/trufflehog_install.py'
+      - 'scripts/check_trufflehog_pin.py'
+      - '.github/workflows/trufflehog-pin.yml'
+  schedule:
+    # Mondays 06:17 UTC — off the top of the hour, which GitHub delays
+    # under load. Note: scheduled runs fire only from the default branch
+    # (so this activates after merge), and GitHub auto-disables crons on
+    # public repos after 60 days with no repository activity. Cron
+    # failure emails go to whoever last edited this file.
+    - cron: '17 6 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: trufflehog-pin-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  pin-integrity:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Fetch upstream checksums.txt (authenticated, retried)
+        env:
+          # Authenticated so we get this repo's GITHUB_TOKEN rate limit
+          # rather than the shared runner-IP unauthenticated limit.
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          PINNED=$(python -c "from clawjournal.redaction.trufflehog_install import PINNED_VERSION; print(PINNED_VERSION)")
+          echo "Pinned TruffleHog version: v$PINNED"
+          for attempt in 1 2 3; do
+            if gh release download "v$PINNED" \
+                 -R trufflesecurity/trufflehog \
+                 -p '*checksums.txt' -O checksums.txt --clobber; then
+              break
+            fi
+            echo "fetch attempt $attempt failed; retrying..." >&2
+            sleep $((attempt * 5))
+          done
+          test -s checksums.txt
+      - name: Compare vendored checksums against upstream
+        run: python scripts/check_trufflehog_pin.py checksums.txt
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -65,8 +65,8 @@ The scan is mandatory. Outcomes:
 Install:
 
 ```bash
-# Any platform — pinned version, sha256-verified against the official
-# release, installed to ~/.clawjournal/bin (preferred over PATH):
+# macOS / Linux / Windows (x86-64 and ARM64) — pinned version, sha256-verified
+# against the official release, installed to ~/.clawjournal/bin (preferred over PATH):
 clawjournal trufflehog install
 
 # Or install it yourself:
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -65,12 +65,16 @@ The scan is mandatory. Outcomes:
 Install:
 
 ```bash
-# macOS
-brew install trufflehog
+# Any platform — pinned version, sha256-verified against the official
+# release, installed to ~/.clawjournal/bin (preferred over PATH):
+clawjournal trufflehog install
 
-# Linux / Windows
-# https://github.com/trufflesecurity/trufflehog#floppy_disk-installation
+# Or install it yourself:
+brew install trufflehog          # macOS
+# Linux / Windows: https://github.com/trufflesecurity/trufflehog#floppy_disk-installation
 ```
+
+The managed copy is downloaded from TruffleHog's own GitHub release artifacts at your explicit request and is only ever invoked as a subprocess. `clawjournal trufflehog status` shows which binary the gate will use.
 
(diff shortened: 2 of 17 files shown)


kai-rayward:
## Round-2 review (fresh lenses) — 9 findings fixed in ca9a534

A second adversarial review pass (deep execution trace, concurrency, failure-mode UX, security, consistency — 56 agents, 3 strict judges per finding) confirmed nine issues the first round missed. All are fixed:

**Bugs**
1. **`already-installed` lied when the gate wouldn't use the binary** — the early exit checked only `target.exists()`, so a non-executable file, a directory, or an off-pin/broken copy reported success (exit 0) while `resolve_binary()` ignored it and the share gate stayed blocked. It now requires `is_file` + `X_OK` + a `--version` match against `PINNED_VERSION`, and repairs anything else with a fresh install.
2. **Verify ran after publish** — `install --force` replaced a working managed binary *before* the `--version` check, so a transient exec failure deleted a known-good binary. The binary is now staged (mkstemp + fsync), verified, and only then `os.replace()`d — a bad download can never displace a working one.
3. **Local I/O errors misreported as `archive-invalid`** — disk-full, a directory squatting on the target, or Windows refusing to replace a running `.exe` all pointed users at the GitHub URL right after the checksum proved the archive was byte-identical to upstream. Failure domains are now split: archive parsing → `archive-invalid`; local write/publish → `install-failed`.
4. **`status --json` exited 0 when the gate binary was missing** while human mode and `install --json` exited 1 — a script trap. Now consistent (JSON still printed before the non-zero exit).

**Improvements**
5. **`windows_arm64` support** — upstream ships the archive; the reviewer extracted it (member is exactly `trufflehog.exe`) and I independently re-verified its sha256 against the release `checksums.txt` before vendoring the row. The test comment claiming "no windows_arm64 release" was factually wrong.
6. **Download progress** — a ~60 MB download was a silent multi-minute hang; human mode now prints a start line (size from `Content-Length`) and a tick every 16 MB.
7. **Actionable failure messages** — `checksum-mismatch` and `download-failed` results carry `hint` text (TLS-intercepting campus/corporate proxies, `selfupdate`, manual fallback) and the CLI prints error + URL + hint.
8. **Version shape normalized** — `status --json` now returns bare `"version": "3.95.5"` (matching `install --json`) plus the raw `"fingerprint"`.
9. **Venv-qualified hints** — `install.sh`/`install.ps1` printed bare `clawjournal trufflehog install` although the venv isn't on `PATH` at that point; they now print `$VENV_BIN/clawjournal …` / `$ClawJournalExe …`. Staged binary is also fsynced before publish, matching `paths.py`'s durability pattern.

Two findings were judged real-but-fine-as-is and intentionally left: the double resolution between `is_available()` and argv build in `scan_file` (benign — wrong block *reason* only, in a vanishing race), and the default-umask bin dir.

Tests: 50 installer tests (was 44); full suite **2225 passed**. Live-verified on this machine: pin-aware `already-installed` runs the real binary, new JSON shape, exit codes.

kai-rayward:
## Design review verdict + drift fix (0391488)

A four-judge design panel (pin-freshness, precedence, lifecycle, and a red team briefed to argue for closing the PR) ruled the architecture **right-with-changes**, unanimously and at high confidence:

- **Pin + vendored checksums: endorsed.** Same-origin checksum fetching adds nothing against the threats that matter (a tampered release or TLS-intercepting campus proxy rewrites archive and checksums.txt alike); floating-latest would let an untested upstream release break the fails-closed gate fleet-wide within a week. Comparable tooling converges on pin + automated bump (pre-commit revs, SHA-pinned Actions + Dependabot), not float.
- **Managed-beats-PATH: endorsed.** Version-dependent resolution would reintroduce the nondeterminism the PR exists to remove; an ambient brew upgrade must not silently change the gate's engine.
- **The blocking gap: silent off-pin drift.** `selfupdate` propagates a `PINNED_VERSION` bump to every machine within hours, but nothing re-installs the binary or even mentions it — and TruffleHog's detectors are compiled in (weekly upstream releases; no separate rules-DB channel like trivy), so a frozen binary erodes exactly the novel-detector coverage a backstop exists for. `status` computed both versions side-by-side and never compared them.

Fixed in 0391488 (warn, never block — an off-pin scanner is still a working gate):
- `managed_off_pin()` — drift signal, managed copy only (a PATH binary's freshness is the package manager's business).
- `trufflehog status` warns on drift with the one-line fix, notes when the managed copy shadows a PATH binary (path shown), and exposes `managed_off_pin` / `shadowed_path_binary` in `--json`.
- events doctor renders "behind pin vX — run `clawjournal trufflehog install`" (additive `off_pin_expected` field).
- **Engine version is now stamped into share artifacts**: `TruffleHogReport.engine` is set from `engine_fingerprint()` at both export chokepoints and persisted in `trufflehog.json` + manifest `redaction_summary` — which scanner version ran is auditable per share.

Also addressed a report that `install` returns `already-installed` for a partial/corrupt/non-executable managed file: that held for f9361ac but was fixed in ca9a534 (the gate requires `is_file` + `X_OK` + `--version` == pin). Verified live — truncated the real managed binary keeping its exec bit, ran the documented command, watched it fall through, re-download, and atomically repair — and locked in with an explicit corrupt-but-executable regression test.

Panel follow-ups deferred to subsequent PRs: post-selfupdate reconciliation of the managed binary, maintainer-side pin-bump automation (scheduled workflow refreshing the six checksum rows), a `CLAWJOURNAL_TRUFFLEHOG_BIN` escape hatch, and a share-path off-pin notice.

Tests: 57 installer tests; full suite **2237 passed**.

PR #81, opened 2026-06-12
clawshare: parallelize on-open failure-value scoring

timyeyuqing:
Follow-up to #78, bringing back parallel on-open scoring so we can compare the speedup while keeping the settle gate and Ctrl-C behavior honest.

## What

Re-parallelizes the on-open failure-value scoring that #78 reverted to sequential:
- The slow AI judge step runs in worker threads, each with its own read connection via `share_flow.score_compute`.
- DB writes stay serial on the main connection through `share_flow.persist_score`, so the index still has one writer while SQLite WAL handles concurrent readers.
- Only settled, unscored traces are scored on open. Active or recent traces remain deferred through `query_unscored_sessions(..., settle_seconds=SCORE_SETTLE_SECONDS)`.
- `workers` defaults to 6 and bounds concurrency by lazily submitting only the active worker window. `--no-score` still skips scoring entirely.

## Haiku is now the default Claude scoring model

Per discussion with Kai, Claude-backed AI features now default to `claude-haiku-4-5`, while Codex continues to default to `gpt-5.4-mini`. This is the shared backend default, so the web daemon, `clawjournal score`, and CLI on-open scoring resolve consistently. `--score-model` and `--model` still override per run.

## Ctrl-C tradeoff

The sequential version in #78 guaranteed that Ctrl-C launched no further judge calls. Parallel keeps at most `workers` calls in flight; on Ctrl-C we stop queueing new work and cancel submitted calls that have not started. In-flight calls may still finish. If strict no-in-flight-work matters more than speed, that remains the tradeoff.

## Local workflow validation

Kai tested this PR branch locally through the ClawJournal workbench and successfully uploaded a package to `https://data.rayward.ai/`.

## Tests

- Local: `python -m pytest tests/ -q` -> 2177 passed
- CI run `27455492513`: smoke + Python 3.10, 3.11, 3.12, and 3.13 all passed

Generated with Claude Code

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ clawjournal set-score <id> --failure-value 4 --failure-evidence "User corrected 
 clawjournal set-score <id> --quality 4               # legacy productivity override
 ```
 
-Scoring uses the current agent's automation CLI by default (`codex exec` in Codex, the Claude CLI in Claude Code); backends are `claude`, `codex`, `hermes`, `openclaw` (override with `--backend`). Claude Code-backed AI features default to `claude-sonnet-4-6`, Codex-backed AI features default to `gpt-5.4-mini`, and the other backends use their own agent defaults unless you pass `--model`. For Codex specifically, run `codex login` or set `CODEX_API_KEY` for headless scoring. The workbench can also auto-score share-ready traces in the background, but only after you confirm a backend once; confirm it headlessly with `clawjournal config --scorer-backend <backend>` (`none` clears it).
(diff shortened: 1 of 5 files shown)


kai-galpha:
Follow-up fix pushed in `707d18a`:

- `score_traces()` now lazily submits only the active worker window, so Ctrl-C stops queue replenishment and does not submit extra queued judge calls.
- README now documents the Claude default as `claude-haiku-4-5`.
- Added regression coverage for interrupt handling and worker-window submission.

Validation:
- Local: `python -m pytest tests/ -q` -> 2177 passed
- CI run `27455492513`: smoke + Python 3.10/3.11/3.12/3.13 all passed

Note: the PR body still has the older test count because the integration cannot update PR metadata (`403 Resource not accessible by integration`).

PR #84, opened 2026-06-16
feat(workbench): add sessions-per-page selector to Inbox

### LINKED ISSUE #85: Inbox: no way to change results-per-page (stuck at 10 + Load more)
## Problem

The session list on the workbench Inbox only displays **10 results at a time**. There's no control to adjust how many items are shown — to see more you have to repeatedly click **Load more**.

## Desired behavior

A selector to choose the number of results per page — e.g. **10, 20, 50, or 100** — so more sessions can be viewed (or selected) at once.

## Notes

- The page size is hard-coded (`const PAGE_SIZE = 10` in `clawjournal/web/frontend/src/views/Inbox.tsx`).
- Related: the **Load more** button stays visible even after the end of the list is reached.

Fixed by #84.

lwshanbd:
Fixes #85.

## Summary

The workbench Inbox listed sessions in fixed batches of 10 with no way to change the page size; the only way to see more was to repeatedly click **Load more**. This adds a **sessions-per-page selector** (10 / 20 / 50 / 100) next to the Sort dropdown.

- Replaces the hard-coded `const PAGE_SIZE = 10` with `pageSize` state (default 10, session-only; not persisted).
- Adds a `<select>` (10/20/50/100) in the toolbar, styled to match the existing Sort select.
- Changing the page size resets the offset and reloads from the top, reusing the existing filter/sort reload pattern.
- Fixes **Load more** pagination with an explicit `hasMore` flag based on a one-row over-fetch (`limit = pageSize + 1`, render only `pageSize`). This hides **Load more** exactly at the end, including exact page-size multiples.
- Guards session-list requests with a latest-request sequence so stale responses from earlier page-size/filter/load-more requests cannot overwrite or append into the current list.
- Advances `offset` only after a successful latest load, so a failed Load more request cannot skip a page.

Frontend-only: `clawjournal/web/frontend/src/views/Inbox.tsx`. The daemon's session-list endpoint passes `limit` through to SQLite, so no backend change is needed. Built `dist/` is gitignored and produced at package time, so it is not committed.

## Test Plan

- [x] `npm run build` (`tsc -b && vite build`) passes; type-checks and bundles.
- [x] `npm run lint` run; fails only on pre-existing repo lint debt unrelated to this PR (`BadgeChip.tsx`, `RedactionReportPanel.tsx`, `SessionDrawer.tsx`, `Toast.tsx`, `Dashboard.tsx`, existing unused vars in `Inbox.tsx`, `SessionDetail.tsx`).
- [x] `git diff --check` passes.
- [x] Daemon smoke: `python -m clawjournal.cli serve --no-browser`; `/`, `/api/features`, `/api/stats`, and `/api/sessions?limit=11&offset=0` respond.
- [x] Manual browser smoke: open Inbox, switch selector to 20 and confirm 20 rows; click **Load more** and confirm 40 rows; switch selector to 50 and confirm it reloads to 50 rows.

### DIFF
--- a/clawjournal/web/frontend/src/views/Inbox.tsx
+++ b/clawjournal/web/frontend/src/views/Inbox.tsx
@@ -188,6 +188,7 @@ export function Inbox() {
   expandedIdRef.current = expandedId;
   const expandedMsgsRef = useRef(expandedMessages);
   expandedMsgsRef.current = expandedMessages;
+  const loadRequestSeqRef = useRef(0);
   // Session ids with an in-flight redacted-content fetch (single-flight guard).
   const loadingIdsRef = useRef<Set<string>>(new Set());
 
@@ -202,6 +203,8 @@ export function Inbox() {
   }, []);
 
   const loadSessions = useCallback(async (currentOffset: number, append: boolean) => {
+    const requestSeq = loadRequestSeqRef.current + 1;
+    loadRequestSeqRef.current = requestSeq;
     setLoading(true);
     const [sortField, sortOrder] = sort.split(':');
     try {
@@ -215,20 +218,29 @@ export function Inbox() {
         failure_mode: modeFilter,
         sort: sortField,
         order: sortOrder,
-        limit: pageSize,
+        limit: pageSize + 1,
         offset: currentOffset,
       });
-      setSessions(prev => append ? [...prev, ...data] : data);
-      // A full batch means there may be more; a short/empty batch is the end.
-      setHasMore(data.length === pageSize);
+      if (requestSeq !== loadRequestSeqRef.current) return;
+      const visibleRows = data.slice(0, pageSize);
+      setSessions(prev => append ? [...prev, ...visibleRows] : visibleRows);
+      setOffset(currentOffset);
+      setHasMore(data.length > pageSize);
     } catch (e) {
+      if (requestSeq !== loadRequestSeqRef.current) return;
       toast(e instanceof Error ? e.message : 'Failed to load sessions', 'error');
     }
-    finally { setLoading(false); setLoaded(true); }
+    finally {
+      if (requestSeq === loadRequestSeqRef.current) {
+        setLoading(false);
+        setLoaded(true);
+      }
+    }
   }, [sort, sourceFilter, projectFilter, typeFilter, recoveryFilter, attributionFilter, modeFilter, pageSize, toast]);
 
   useEffect(() => {
     setOffset(0);
+    setHasMore(false);
     setSelectedIds(new Set());
     setFocusIndex(-1);
     loadSessions(0, false);
@@ -805,7 +817,7 @@ export function Inbox() {
       {hasMore && sessions.length > 0 && (
         <div style={{ textAlign: 'center', marginTop: '8px' }}>
           <button
-            onClick={() => { const n = offset + pageSize; setOffset(n); loadSessions(n, true); }}
+            onClick={() => { void loadSessions(offset + pageSize, true); }}
             disabled={loading}
             style={{
               padding: '5px 18px', background: colors.gray100, color: colors.gray700, border: `1px solid ${colors.gray300}`,
--- a/clawjournal/web/frontend/src/views/Inbox.tsx
+++ b/clawjournal/web/frontend/src/views/Inbox.tsx
@@ -10,7 +10,6 @@ import { GettingStartedGuide } from '../components/GettingStartedGuide.tsx';
 import { ZeroState } from '../components/ZeroState.tsx';
 import { colors, selectStyle, btnPrimary, btnDanger, btnSecondary } from '../theme.ts';
 
-const PAGE_SIZE = 10;
 // Show only the top few task-type chips by default; the long tail collapses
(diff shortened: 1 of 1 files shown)


lwshanbd:
### Claude Code review

**PR #84** — `feat(workbench): add sessions-per-page selector to Inbox` · single file, +20/−6 (`clawjournal/web/frontend/src/views/Inbox.tsx`)

Reviewed by 4 specialist agents (code-review, silent-failure, comments, test-coverage). Type-design review skipped — no new types, just two primitive `useState` hooks. Overall this is a clean, well-scoped change; the `hasMore` rewrite (`data.length === pageSize`) is a genuine correctness improvement over the old accumulating `sessions.length >= PAGE_SIZE` heuristic, which stayed visible forever after the first full page.

#### Critical
None.

#### Important
1. **Load-more failure silently skips a page** — `Inbox.tsx:802` + `:221-227`
   The Load-more handler advances `offset` *before* the fetch resolves:
   ```js
   onClick={() => { const n = offset + pageSize; setOffset(n); loadSessions(n, true); }}
   ```
   On fetch error the `catch` only toasts — `setSessions` (in `try`) never runs, so no rows are appended, but `offset` stays advanced. The next successful click computes `offset + pageSize` again, **permanently skipping the failed page** with no indication to the user. Fix: advance `offset` only on success (inside `try` after `setSessions`), or roll it back in the `catch`.

2. **Stale-response race when changing `pageSize` (or filters) mid-flight** — `Inbox.tsx:204-228`, `:802`
   `loadSessions` has no generation guard or `AbortController`. Click Load-more (offset 10, limit 10) → before it resolves, switch page size to 50 → the reload effect fires `loadSessions(0, false)`. If the first (append) request resolves last, its `setSessions(prev => [...prev, ...data])` appends a stale offset-10/limit-10 slice onto the fresh 50-row replace → duplicated/overlapping rows and a wrong `hasMore`. The page-size selector makes this a one-click trigger. The file already uses a single-flight pattern elsewhere (`loadingIdsRef` in `handleExpand`) — apply the same idea: capture a request-id ref at call start and bail before `setSessions`/`setHasMore` if it's stale.

#### Suggestions
- **Redundant `sessions.length > 0` guard** — `Inbox.tsx:802`: `{hasMore && sessions.length > 0 && ...}` — `hasMore` is only `true` after a full batch loaded, which already implies a non-empty list. Either drop the clause or add a one-line comment naming the edge case it protects (e.g. the brief window between a filter change and the refetch).
- **Initial-load failure leaves stale `hasMore`/`sessions`** — `Inbox.tsx:223`, `:230-236`: on a failed reload, `setSessions`/`setHasMore` don't run, so the prior filter's list and flag persist (a toast does fire). Internally consistent, but if you want an unambiguous failure state, reset them at the top of the effect where you already reset `offset`/`selectedIds`/`focusIndex`. Optional.

#### Strengths
- `hasMore` heuristic is correct and fixes the long-standing "Load more never disappears" bug.
- `pageSize` correctly added to both the `loadSessions` `useCallback` deps and the reload `useEffect` deps — no stale-closure bug (the `react-hooks` lint rule would have caught a miss).
- The new comment at `:222` explains *why* (`A full batch means there may be more…`) rather than restating the code — a good "why" comment, and removing `PAGE_SIZE` orphaned no other comments.

#### Tests
No coverage gap to flag. The repo has **zero frontend test infrastructure by design** — no runner/`vitest`/`testing-library` in `package.json`, no spec files under `src/`, and CI treats the frontend as a build artifact (`npm run build` + wheel-contains-`dist/index.html` smoke check, per `CLAUDE.md`). The PR's `build` + `lint` + manual test plan matches every other frontend change. Adding a harness for a 4-option dropdown would be disproportionate and is a separate repo-wide decision.

#### Recommendation
Not blocking, but **fix the Load-more offset-skip (Important #1)** before merge — it's a real data-visibility bug. The race (Important #2) is worth addressing while you're in here. Everything else is optional polish.

🤖 Generated by `pr-dual-review` skill — Claude side

lwshanbd:
### Codex review

*Independent second opinion — generated without sight of the Claude-side findings.*

### Summary
Overall risk is low for the feature itself, but the new page-size control exposes an existing async pagination race more often. I found one user-visible correctness issue around stale responses corrupting the inbox list.

### Critical (must fix)
None

### Important (should fix)
- `clawjournal/web/frontend/src/views/Inbox.tsx:221` — stale `loadSessions` responses can overwrite or append into the list after `pageSize` changes — changing page size while an earlier initial/load-more request is in flight can leave the UI showing the wrong batch size or duplicate/mixed-offset sessions — add a request generation ref or abort controller and guard `setSessions`, `setHasMore`, and `setLoading` so only the latest request updates state.

### Nits / suggestions
- `clawjournal/web/frontend/src/views/Inbox.tsx:223` — `hasMore = data.length === pageSize` still shows one extra "Load more" click when the total is an exact multiple of the page size; over-fetch `pageSize + 1` and render only `pageSize` rows if you want exact has-more behavior.

### Questions for the author
None

🤖 Generated by `pr-dual-review` skill — Codex side (independent second opinion)

kai-rayward:
Thanks @lwshanbd !

PR #110, opened 2026-07-11
feat(skills): complete self-improving lessons workflow

JustMeStudio:
## Summary

Builds on #108’s self-improving-skills design with implementation hardening, Windows compatibility fixes, and an automated Codex installation-loop test.

### Active lesson set

- Caps the active installed lesson set at 5 rules.
- Keeps additional historical rules in the local database as proposed, dropped, or rejected instead of deleting them.
- Updates merge behavior and tests to preserve a mix of `do` and `avoid` lessons within the five-rule context budget.
- Aligns the design docs with the five-active-rule implementation.

### Codex and Windows compatibility

- Uses the full path returned by `shutil.which()` when launching agent backends.
- Fixes Windows npm shim handling such as `codex.cmd`.
- Explicitly decodes agent subprocess output as UTF-8 with replacement for invalid bytes instead of relying on the Windows GBK code page.
- Reads scoring rubrics and session blobs explicitly as UTF-8.
- Improves missing-file/command errors during scoring.
- Honors the `HOME` environment variable for Claude and Codex global installation paths, while falling back to `Path.home()`.

### Install workflow

- Keeps the generated active set limited to the five highest-ranked lessons.
- Installs the Codex lesson region into the global `.codex/AGENTS.md` surface.
- Records successfully installed rules as `kept` with an `installed_at` timestamp.
- Preserves dropped rules in local history.
- Keeps repeated Codex installation idempotent so the managed `AGENTS.md` region is replaced rather than duplicated.

### Tests

Adds a CLI closed-loop installation test covering:

1. A generated lesson result.
2. Interactive user confirmation with `y`.
3. Codex-only installation into a temporary `HOME`.
4. Creation and content validation of `.codex/AGENTS.md`.
5. Confirmation that no Claude skill is written for a Codex-only target.
6. Database persistence as `kept` with `installed_at`.
7. A second installation without duplicating the managed region.

Also adds or updates tests for:

- the five-rule active cap;
- preservation of the `do` / `avoid` mix;
- archival of dropped rules;
- resolved Windows command paths;
- Windows `.cmd` shims;
- UTF-8 subprocess decoding;
- UTF-8 scoring file reads.

## Documentation

- Adds implementation review notes covering explicit user-triggered distillation, active-set versus archive behavior, native agent memory boundaries, and correction-signal reliability.
- Updates the plan, validation design, and competitor comparison to describe the five-rule active context budget.

## Validation

- `pytest tests/skill/test_cli_skill.py -q` — 6 passed
- `pytest tests/skill/ -q` — 143 passed
- `pytest tests/scoring/test_backends.py tests/scoring/test_scoring.py` — 154 passed

## Manual Validation

Manually exercised:

- Codex CLI discovery through a Windows npm installation.
- UTF-8 scoring on a Chinese Windows environment.
- Session indexing from Claude and Codex sources.
- Codex scoring through `codex exec`.
- Preview behavior for `clawjournal skill`.
- TruffleHog and project-confirmation preflight gates.

## Notes

- CI tests use a fake distillation backend and do not make real Claude or Codex calls.
- The closed-loop test validates the workflow after a valid rule set has been produced.
- Large-session scoring timeouts and richer distillation diagnostics remain separate follow-up work.


kai-rayward:
@JustMeStudio curious curious how would you like to test these skills? And how to tell these skills may be useful?

kai-rayward:
## Review

**Verdict: hold for now.** There's a lot of genuinely valuable work in here, but three things need resolving before merge, plus one bug.

### 1. CI never ran

The workflow run for the head commit is stuck at `action_required` (fork-PR approval gate), which is why the PR shows no checks. It needs a maintainer "Approve and run" and a green matrix before anything else.

### 2. `MAX_INSTALLED_RULES` 10→5 reverses #96

#96 deliberately widened the installed set to 10 so lessons accumulate across weekly runs instead of churning. The counter-model here (≤5 active in agent context, larger local archive of proposed/dropped/rejected states) is coherent and the archive test backs it up — but it's a product decision being reversed inside a PR titled as review notes, stacked under a docs PR. Let's make that call explicitly rather than have it ride along; it shouldn't be implicit in a merge.

### 3. Scope: the best fixes are stuck on a docs branch

The most valuable, least controversial changes are runtime fixes unrelated to the skills design docs:

- `require_backend_command` returning the `shutil.which()`-resolved path (makes `codex.cmd` npm shims work on Windows)
- UTF-8 + `errors="replace"` on all agent subprocess decoding (fixes GBK-codepage breakage) — I checked all five `subprocess.run` sites in `backends.py`; none missed
- explicit UTF-8 file reads in scoring
- the `FileNotFoundError` handler in `_score_single_session`

These follow the exact precedent of #93/#94/#112, which went straight to `main`. On this branch they only reach `main` when the whole #108 docs stack merges. Please split them into a `main`-targeted PR so Windows users get them without waiting on the design docs.

### 4. Bug: `_home_dir()` is wrong on the platform it targets

`clawjournal/skill/install.py` now prefers `$HOME` over `Path.home()`. On POSIX that's a no-op (`Path.home()` already reads `HOME`). On Windows it's a behavior change in the wrong direction: Python's `Path.home()` uses `USERPROFILE` and ignores `HOME` — and so do Node's `os.homedir()` (Claude Code) and Codex's home resolution. Git Bash/MSYS users commonly have `HOME` set to a *different* path, so lessons would install to `$HOME/.codex/AGENTS.md` where Codex never reads them — a silent no-op install.

Since the helper exists to make the new test's `monkeypatch.setenv("HOME", ...)` work on Windows dev machines, better options: monkeypatch `Path.home()` in the test, or use a dedicated `CLAWJOURNAL_HOME`-style override. Production resolution should stay `Path.home()`.

### What's great and should stay

- The closed-loop Codex install test: interactive confirm → `AGENTS.md` write → DB `kept` + `installed_at` → idempotent re-install without duplicating the managed region. Exactly the coverage this workflow needed.
- The confirm prompt naming only the targets actually being installed.
- Docs kept consistent with the cap change (competitor-comparison, plan, validation-design B6).
- Nit, no action needed: `_format_install_targets` maps `workbuddy`, but `--target` argparse still only accepts `claude`/`codex` — harmless forward-provisioning.

Suggested path: approve CI, split item 3 into a `main` PR, fix `_home_dir()`, and settle the 10→5 question explicitly; the review-notes doc plus whatever cap decision lands can then merge cleanly into the #108 stack.

kai-rayward:
Pushed `7a19ea9` addressing item 4 of the review: agent install paths now resolve via `Path.home()` again. On Windows, Claude Code and Codex read their global dirs from `USERPROFILE` (what `Path.home()` uses), so preferring `$HOME` would have sent lessons to a path the agents never load for Git Bash/MSYS users. The closed-loop test now sets both `HOME` and `USERPROFILE` so it stays hermetic on every platform. CI approved and green (3.10–3.13 + smoke); `tests/skill/` + `tests/scoring/test_backends.py` (238 tests) pass locally. Merging with the ≤5 active-rule cap accepted.

PR #111, opened 2026-07-11
Detect new activity in submitted traces

richardopenrefinery:
## Summary

- fingerprint parsed trace content during indexing
- record the exact revision included in an export and promote it after a successful upload
- make continued sessions share-ready again when their messages change
- preserve safe behavior for unchanged and legacy submissions
- report changed existing traces from `clawjournal scan`

## Root cause

The share queue permanently excluded any session ID associated with an uploaded share. Agent sessions can continue growing under the same session ID, so newly appended signals were indexed but remained hidden from the queue.

## Impact

Users can submit a trace, continue working in the same agent session, scan again, and package the updated trace in a new share. Routine rescans of unchanged traces remain idempotent.

## Validation

- `pytest tests/workbench/test_daemon.py::TestScanner tests/workbench/test_index.py::TestShares -q` — 26 passed
- `python -m compileall -q clawjournal`
- `git diff --check`

### DIFF
--- a/clawjournal/cli.py
+++ b/clawjournal/cli.py
@@ -1238,6 +1238,7 @@ def _run_scan(source_filter: str | None = None) -> None:
     results = scanner.scan_once()
 
     total_new = sum(results.values())
+    updated = scanner.last_updated_count
     if total_new:
         print(f"Indexed {total_new} new sessions:")
         for source, count in sorted(results.items()):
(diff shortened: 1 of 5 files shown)


richardopenrefinery:
Add the detector for such traces: 
"If I continue working on the same project in the same trace (very common in science because the work is usually long-horizon and iterative), when I run clawjournal scan again, it does not detect the “new trace.”

PR #114, opened 2026-07-15
feat(sharing): add consented weekly automatic uploads

### LINKED ISSUE #113: Share: make weekly uploads automatic after one-time consent
## Goal

After one successful hosted manual share, let a participant authorize ClawJournal to share future eligible traces from an exact source/project scope without reviewing each bundle.

This issue is the replacement V1 contract. The detailed implementation plan lives locally at docs/issue-113-auto-weekly-upload-plan.md.

## Fixed V1 decisions

1. The first automatic cycle is due seven days after enrollment. Later successful or nothing-new cycles reset the seven-day clock and run on the next supported Claude Code or Codex SessionStart.
2. Each scheduled or explicit Run now cycle selects at most five traces, ordered by stored failure-value score with a deterministic fallback. Automatic runs never score synchronously.
3. Run now is an explicit extra capped cycle and resets the next due time. V1 does not enforce a server-side weekly quota across explicit Run now calls.
4. Missed weeks produce one catch-up cycle, not a burst.
5. Enrollment is future-only. Pre-enrollment history never drains automatically.
6. Updated revisions require the existing fresh-approval gate.
7. Scheduling uses Claude Code and Codex SessionStart hooks, not cron, launchd, systemd, or a daemon timer.
8. The client and hosted recurring-upload capability ship together. Manual sharing remains unchanged.

These decisions intentionally replace the earlier “all eligible/no ranking” text.

## Recurring authorization

The current manual-share consent is bundle-specific and cannot authorize unseen future bundles. Enrollment requires separately versioned, privacy-approved recurring scope authorization that discloses:

- future selected traces may upload without per-bundle review;
- exact local source/project scope;
- future-only boundary, top-five cycle cap, activity-triggered cadence, and Run now behavior;
- local anonymization/redaction/findings gates and optional AI-PII provider/usage;
- 24 hours of unchanged content as the completion rule for append-only sources without a trusted close marker;
- preview, hold, pause, disable, retention, and in-flight request behavior; and
- participant authorization to share traces from the enrolled scope.

New projects remain out of scope until explicitly added. Any redaction/allowlist, AI provider, source/project, or destination-origin profile change blocks unattended AI/upload until the participant reviews the current profile and terms.

## Eligibility and selection

One shared candidate service powers status, preview, and run.

An automatic revision must:

- be in the exact enrolled source/project snapshot with current project confirmation;
- have end_time after server-accepted enrollment time;
- satisfy a tested explicit-close rule or remain at the same content revision for at least 24 hours with end_time at least 24 hours old;
- be new, shortlisted, or approved, never blocked/segmented;
- have effective hold state auto_redacted or released;
- have its raw blob;
- not already be shared at the same content revision;
- pass the changed-revision approval gate; and
- still match its content revision at package and egress.

Order by scored before unscored, failure-value descending, end_time descending, then session_id. Take up to five. A conservative pre-AI size check may defer only a ranked suffix.

## Safety and recovery

Before public enablement:

- fix the finalized AI-PII coverage schema so AI enabled means zero rules-only coverage for fresh and cached artifacts;
- move the complete release/hold/revision gate to the package boundary before AI and repeat it immediately before egress;
- make scans strict for enrolled sources; partial source/parser failures cannot upload or advance cadence;
- persist exact source/project scope locally;
- seal, fsync, and SHA-256 the exact ZIP bytes before the first request;
- persist client submission identity and included revisions before egress;
- reconcile ambiguous submissions by receipt lookup and retry only identical bytes under the identical id;
- preserve both mandatory TruffleHog passes;
- use one SQLite authority, a generation counter, and an OS-released run lock;
- reuse shares/share_sessions and existing pending_review hold/history rather than parallel baseline/quarantine systems.

Mapped findings place exact revisions in pending_review, stop that package, and allow a later fresh cycle to consider remaining safe revisions. Unmappable findings fail closed.

The atomic submitting transition is the honest in-flight boundary. Pause/disable before it wins; afterward status reports that one request may finish.

## Hosted contract

Capability discovery advertises recurring_upload_api_version 1, max five sessions, max bytes, exact receipt lookup, supported client versions, and duplicate-revision enforcement.

Minimal operations:

1. Fetch recurring authorization and retention versions.
2. Idempotently create an enrollment using verified identity plus a durable client_enrollment_id.
3. Read/update enrollment status and scope/terms.
4. Revoke idempotently.
5. Submit an exact ZIP using active enrollment authorization, client_submission_id, authorization revision, and pseudonymous revision keys.
6. Look up a receipt by client submission ID.

The server issues two opaque credentials:

- active: status/update/upload;
- recovery-only: revoke and receipt lookup, never upload.

The server hashes credentials, computes the received ZIP hash, enforces at most five sessions, rejects duplicate revision keys for the same participant/study and stable installation across re-enrollment, and returns the prior receipt for the same submission id/exact bytes. Same id/different bytes is a conflict.

Manual one-shot credentials are never an unattended fallback.

## User surface

After the first successful hosted share, DoneStep may offer automatic weekly sharing when the capability is available.

Settings and equivalent CLI commands provide:

1. enable/review scope and terms;
2. read-only status;
3. preview, optionally after a strict local refresh;
4. Run now;
5. pause/resume;
6. disable.

Status distinguishes mode (Off/On/Paused), health (Ready/Action required/Retrying), Running, and Revocation pending. Findings review remains in the Share workflow.

Hooks must preserve unrelated/OpenRefinery configuration, use PascalCase SessionStart, exit quickly, and spawn detached work only when due.

## Failure behavior

- nothing_new: quiet success and cadence advances;
- retrying: strict scan, network, AI, 429, or retryable server failure; enrollment stays enabled;
- action_required: scope/terms/auth incompatibility, missing hook, unsupported source, unmappable findings, or residual oversized bundle;
- ambiguous: preserve and reconcile exact artifact/key before new selection.

Disable takes local effect first, removes hooks and active upload authority, and retains only recovery authority until revocation/receipts are definite.

## Delivery

1. Shared gate and exact-artifact hardening.
2. Local schema, candidate service, credentials, runner/recovery, and tests behind unavailable/default-off gating.
3. Claude/Codex hooks, CLI, and local daemon API.
4. Hosted recurring enrollment/idempotency/dedupe/receipt contract and compatibility/staging tests.
5. DoneStep/Settings UI, rebuilt assets, README/PRIVACY/ARCHITECTURE updates, and capability canary.

No UI may report On before server enrollment plus protected local credential/state persistence are complete.

## Acceptance criteria

1. No successful hosted manual share means no enrollment.
2. Recurring authorization is distinct, versioned, server-recorded, privacy-approved, and accurately discloses unattended future upload and optional AI processing.
3. No pre-enrollment, incomplete, unsupported, blocked, held, out-of-scope, stale, missing, or duplicate revision uploads.
4. Each cycle selects at most five deterministically and never scores synchronously.
5. Partial scans, profile/terms drift, or failed gates cannot upload or advance cadence.
6. Pause/disable/hold/revision/generation changes win before submitting.
7. Ambiguous submissions reconcile to at most one receipt using identical bytes/id.
8. Server duplicate, count, authorization-revision, and idempotency conflicts are transactional.
9. AI-PII enabled means zero rules-only coverage; anonymization is pre-AI, deterministic redaction is independent, and both TruffleHog gates remain.
10. Missing capability never produces On or one-shot-token fallback.
11. Preview and Run share one candidate contract and explain exclusions.
12. Disable removes local upload authority immediately and preserves only bounded recovery authority.
13. Claude/Codex hooks preserve unrelated configuration and never block agent startup.
14. Manual workbench, CLI, ZIP, hold, and one-shot Share behavior is unchanged.
15. Backend, frontend/build, wheel smoke, hosted compatibility, staging real-gate, and documentation checks pass in the final merged state.

## Non-goals

- exact wall-clock scheduling or OS schedulers;
- automatic pre-enrollment history;
- hard server weekly quota across explicit Run now calls;
- post-package partial repackaging;
- synchronous scoring;
- hooks beyond Claude Code/Codex;
- self-hosted/arbitrary automatic destinations;
- incentives/rewards;
- cross-install dedupe after local state/hash-salt loss;
- recalling a request past submitting;
- deleting prior uploads on disable; or
- replacing manual Share.

JustMeStudio:
Closes #113

## Summary

Implements the ClawJournal client side of recurring weekly data sharing after one-time consent.

## What changed

- Adds versioned recurring enrollment with explicit source/project scope
- Triggers due runs from Claude Code and Codex SessionStart hooks
- Selects at most five eligible traces using deterministic ordering
- Enforces future-only, 24-hour stability, release, revision, PII, and secret gates
- Seals and hashes the exact ZIP before submission
- Preserves exact bytes and submission IDs for safe retries
- Adds receipt reconciliation for ambiguous responses
- Adds status, preview, run, pause, resume, and disable controls
- Adds Settings and post-share enrollment UI
- Preserves unrelated agent hooks

## Validation

- Real Codex JSONL → production Scanner → SQLite → selection → privacy gates
  → real TruffleHog → bundle → ZIP/fsync/SHA-256 → persisted submission metadata
- Only the final hosted-server request is mocked
- Automatic upload tests: 8 passed
- SessionStart hook tests: 3 passed
- Daemon automatic-upload API tests: 2 passed
- Frontend production build passed
- Wheel smoke build passed
- `git diff --check` passed

## Remaining external work

The hosted recurring-enrollment, submission, receipt-reconciliation, and revoke
endpoints must be implemented and deployed separately. A staging canary is still
required before production rollout.


kai-rayward:
@JustMeStudio Great work!
I'm curious on how would you like to test such a big changes carefully?

kai-rayward (changes requested):
**Verdict: the architecture and invariant discipline are strong, but there is one confirmed functional bug that breaks the PR's flagship flow — the post-share "Enable weekly sharing" button can never succeed — plus a few robustness gaps in the retry machinery. Requesting changes for the enrollment-token issue and the artifact-hash check; the rest can be follow-ups.**

Validation on my side: checked out the PR head locally; all 14 new tests plus the full suite pass (2016 passed; the one failure, `tests/test_selfupdate.py` contention test, is a pre-existing flake that also passes standalone on the merge-base), CI is green, and the SessionStart hook command measures ~0.1s — negligible per-session overhead.

## Confirmed bug (blocks the primary flow)

**Enabling auto-upload after a manual share always fails.** `enable_enrollment` requires `config["verified_email_token"]` (`clawjournal/auto_upload.py:162-164`), but every successful manual hosted upload deletes that token — `_clear_stored_upload_token()` in `submit_share_to_hosted`, and `test_share_success_clears_cached_upload_token` asserts exactly this. Since enrollment *also* requires a completed hosted share, at the moment the DoneStep "Enable weekly sharing" button appears the token is guaranteed gone, and the daemon handler passes no `authorization_token`. The user gets "Fresh verified identity is required…" with no path forward; Settings/CLI enable only work in the odd window where someone verified email but didn't submit. The enable flow needs to drive a fresh email verification (reuse the existing verify step), or the manual-share receipt should carry a separate short-lived enrollment grant. The tests inject the token directly, which is why nothing caught this; an end-to-end "share → enable" test would have.

## Should fix before merge

- **Retry never verifies the sealed artifact's hash.** The pending-share path selects `artifact_sha256` but only checks `os.path.isfile` (`clawjournal/auto_upload.py:653-667`), and `submit_share_to_hosted` reads `artifact_path` blind. A truncated or modified file uploads as-is, despite the error message promising "refusing a non-identical retry" and the stated invariant that recovery reuses exact bytes. Recompute the SHA-256 and route mismatches to `review_pending_artifact`.
- **A crash mid-run leaves state stuck.** Nothing catches unexpected exceptions between `_record_run` and `_finish_run` — e.g. `json.loads(pending["revisions_json"])` raises `TypeError` if that column is NULL, or `_seal_artifact` hits an `OSError`. The detached worker's output goes to DEVNULL, so the failure is silent, the run row stays `running` forever, and status reports `running: true` indefinitely. Wrap the body so the run row is always finalized.

## Design concerns (fine as follow-ups, worth deciding now)

- **No retry backoff.** On failure, `next_due_at` doesn't advance, so *every* Claude/Codex session start spawns a full cycle — forced capabilities fetch, full source scan, package, upload attempt — until the problem clears. The file lock prevents overlap but not frequency. A persisted `next_attempt_at` or per-due-window attempt cap would bound this.
- **`GET /api/auto-upload/status` phones home.** `enrollment_status` calls `_fetch_hosted_share_capabilities()` even for never-enrolled users, and both Settings and Share mount it — so opening Settings now contacts the hosted API (cached 300s). Given this repo's history of gating daemon egress (#107), consider cache-only or enrolled-only discovery here.
- **`clawjournal config` prints the new long-lived tokens.** `_mask_config_for_display` masks only `redact_strings`, so `recurring_upload_token` and the recovery token print in plaintext. These are standing credentials, unlike the short-lived one-shot token; mask them. (File perms are fine — `mkstemp` gives config.json 0600.)
- **Env override beats the explicit `home=` parameter in the scheduler.** `_path()` checks `CLAUDE_CONFIG_DIR`/`CODEX_HOME` first (`clawjournal/auto_upload_scheduler.py:21-26`), so anyone running pytest with those set would have the test suite write a hook into their *real* settings.json. Make `home=` win, or `monkeypatch.delenv` in the tests.

## Smaller items

- PRIVACY.md/README promise a "tested close marker" exception to the 24-hour stability rule; no close-marker logic exists in `select_pending_sessions` — the rule is unconditional. Drop the claim or implement it.
- The ambiguous-reconcile path (`clawjournal/auto_upload.py:773-786`) marks the share `shared` but skips `bundle_hash`/`hosted_submission_url` that the normal success path records, losing audit fields.
- Once disabled with a failed revocation (`revocation_pending=1`), the Settings UI hides the Disable button, so only re-running CLI `disable` retries revocation.
- `from .config import CONFIG_DIR` binds the value at import, which sidesteps the autouse `tmp_config` guard — the real-pipeline test has to re-patch `auto_upload.CONFIG_DIR` by hand, and future tests that forget will write to the real home. Same for the hardcoded `RUN_LOCK` path (duplicates `CONFIG_DIR`).
- `auto_upload_runs` grows unboundedly; `_ACTIVE_STATES` includes `"off"` despite its name; `deferred_count` counts never-eligible sessions, overstating what's "deferred"; consent via `window.confirm` with a wall of text is much weaker UX than the manual flow's consent step for what is a *standing* authorization.

## What's genuinely good

The key invariants are respected with real depth: `package()` re-runs release/source/revision/already-shared gates, and `submit_share_to_hosted` repeats all four *plus* the enrollment-generation check immediately before egress, so pause/disable reliably wins before the submitting boundary. Selection is snapshot-scoped (new projects never drain), future-only against the server-accepted timestamp, and the ≤5 cap is enforced in both selection and submit. The same `client_submission_id` is correctly reused across retry attempts. Hook install/remove preserves unrelated hooks and is marker-scoped, with tests. The manual share path explicitly refuses server-offered recurring tokens — a nice adversarial test. Migration is additive v8 per convention, HTTPS is enforced, and all new endpoints sit behind the bearer token.

https://claude.ai/code/session_01PcJsydoENPknWvoFE2mWvX

kai-rayward:
@JustMeStudio — first, thank you for this. It's a serious, carefully-built PR: CI is green, the tests you wrote pass cleanly in an isolated checkout, and the parts of the contract you implemented show real invariant discipline — the release/revision gates re-run right before egress, retries reuse the exact sealed bytes and client submission ID, hook install/remove is marker-scoped and preserves unrelated hooks, and the test asserting that a manual share refuses server-offered recurring tokens is exactly the kind of adversarial thinking this feature needs. The review above stands as the detailed technical record, but none of it takes away from the quality of the attempt.

I want to be transparent about where this lands, though: **we're going to close this PR as superseded** rather than iterate on it. Here's the full reasoning.

## A parallel implementation exists

While this PR was in flight, a parallel implementation of the recurring-upload contract (#113) has been under heavy development in-tree, and it has grown into the version we need to ship. It's roughly four times the size of this PR's core and resolves every blocking or major item from the review by construction, not by patching:

- **The enrollment dead-end** (the one-shot `verified_email_token` being consumed by the manual share that gates enrollment) is solved by a typed `email_verification_required` response that both the CLI and a dedicated authorization dialog handle by driving a fresh inline email verification. Enrollment *updates* reuse the stored active credential instead of the email token entirely.
- **Exact-bytes integrity on retry** goes further than re-hashing the ZIP: the sealed artifact's SHA-256 is re-verified (constant-time) both when the ledger is read and again immediately before the POST, and the *raw parser inputs* are fingerprinted (`dev/ino/size/mtime/sha256` with torn-read detection) and re-validated so a retry refuses to proceed if the source files changed after sealing.
- **Crash-safety and pacing**: the runner records exponential backoff (`next_retry_at`, honored by the SessionStart due check), "running" is a display overlay rather than authority (the OS lock is the authority, and orphaned overlays are cleared on entry), and receipt + cadence checkpointing is a single crash-atomic transaction.
- **No phone-home from status**: `GET /api/auto-upload/status` is network-free by contract (read-only SQLite, no migrations); capability availability for the post-share offer is a config bit cached at manual-share time, and Enable revalidates live.
- **Credentials never touch config.json**: a purpose-separated credential file (0700/0600 enforced fail-loud, origin-pinned, atomic writes) with distinct active/recovery authority — disable keeps only a recovery tombstone that can reconcile receipts but can never upload.
- **Consent is a versioned challenge, not a confirm dialog**: the server defines the authorization text and versions; acceptance must echo the exact versions, so no client can blanket-`--yes` its way past updated terms.
- Plus a set of things this PR had no occasion to build: per-AI-call control gates (a pause during PII review call 1 is seen before call 2), an egress-profile hash that also covers redaction-policy drift (allowlist/custom-string changes pause the enrollment), strict-scan log hygiene so unattended runs can't leak content into logs, and frontend test infrastructure (vitest + component tests for the consent, polling-race, and revocation-retry flows).

## Why we can't merge both

The forcing function is mechanical, not just preference: both implementations stamp `PRAGMA user_version` 7/8 with **structurally different** `auto_upload_enrollment` shapes. A database touched by one is silently mis-migrated by the other — there is no additive merge path, and renumbering/reconciling this PR against the other schema would effectively mean rewriting it.

## What carries forward from this PR

Closing the PR doesn't discard the work:

- The review findings it surfaced (and your responses to the contract) sharpened the successor — several of its tests exist specifically because this PR proved those paths needed pinning down.
- The **manual-share-must-refuse-recurring-tokens** test pattern is adopted in spirit as a standing invariant.
- A configurable cadence (`--cadence-days`) is a genuinely nice knob this PR had and the successor deliberately hardcodes to the challenge-fixed 7 days for V1 — if we loosen that post-V1, this PR is the reference.

The successor branch will land shortly and will be linked here; I'd genuinely value your eyes on it — you know this contract as well as anyone now, and follow-up contributions (especially around #113's remaining hosted-side work and the staging canary) are very welcome.

Thanks again for the effort and rigor here, Dean.

https://claude.ai/code/session_01PcJsydoENPknWvoFE2mWvX

kai-rayward:
The successor implementation is now up as #115 — closing this as superseded, per the rationale above. Thanks again @JustMeStudio; your review eyes on #115 would be very welcome.

PR #115, opened 2026-07-16
feat(sharing): consented weekly automatic uploads (V1)

### LINKED ISSUE #113: Share: make weekly uploads automatic after one-time consent

kai-rayward:
Closes #113. Supersedes #114 (see the [review](https://github.com/rayward-external/clawjournal/pull/114#pullrequestreview-4710696357) and [supersession rationale](https://github.com/rayward-external/clawjournal/pull/114#issuecomment-4990674221) there for the full comparison).

## Summary

Implements the ClawJournal client side of recurring weekly data sharing after one-time, versioned consent. Default-off, future-only, capped at five traces per cycle, exact-scope, and unavailable without hosted protocol-v1 discovery plus a successful manual receipt.

## What changed

**Enrollment & identity**
- Enrollment is a transactional state machine: a server-defined authorization challenge bound to exact consent/retention versions (no `--yes` bypass), fresh email verification for new enrollments (the one-shot manual token is consumed and its removal verified), CAS on generation, hook-file snapshot/rollback, and a compensating revoke if local persistence fails after server create.
- Credentials live in a purpose-separated, origin-pinned, fail-loud 0700/0600 store — never `config.json`. Disable removes active authority first and keeps only a recovery tombstone that can reconcile receipts but can never upload. `clawjournal config` masks all token-like keys recursively.

**Scheduling & runner**
- Claude Code and Codex `SessionStart` hooks run a bounded, fail-open, network-free due check (read-only DB, no migrations) and detach a one-shot runner under an OS-released lock. Failing cycles back off exponentially (15 min × 2ⁿ, capped at 24 h, honoring server `retry_after`); hard crashes stamp the same backoff. No cron or daemon timer — a missed week is one capped catch-up.

**Selection & egress integrity**
- Selection is snapshot-scoped, future-only against the enrollment boundary, ordered by stored failure-value scores (never scoring synchronously), and gated on 24-hour revision stability, hold state, exact-revision dedup across all successful shares, fresh-approval for changed traces, and raw-source availability. Every exclusion is counted by reason.
- Egress reuses exact sealed bytes: the ZIP is fsynced and SHA-256-verified (constant-time) before every retry; raw parser inputs are fingerprinted (`dev/ino/size/mtime/sha256`, torn-read detection) and re-validated; release/revision/profile/generation gates repeat at selection, **before every AI call**, after packaging, and under a cross-process lock at the atomic `sealed → submitting` transition. Config/allowlist profile changes pause an active enrollment.
- Ambiguous responses reconcile by client submission ID using recovery-only authority on fixed endpoints that survive public discovery going dark. AI-PII coverage failures block automatic bundles fail-closed; both TruffleHog gates still apply.

**Workbench & CLI**
- Versioned-terms authorization dialog (explicit checkbox, inline email re-verification, agent selection, scope summary), status panel with per-agent hook diagnostics, per-reason eligibility counts, preview, Run now, pause/resume, revocation retry, and a post-share offer with per-receipt dismissal. `GET /api/auto-upload/status` and preview are network-free by contract.
- `clawjournal auto-upload enable|status|preview|run|pause|resume|disable` with interactive terms acceptance and fresh-verification flow; `--json` for scripting.
- Strict scans keep transcript content out of daemon logs via an allowlisted log guard.

## Validation

- Full suite: **2,664 passed** (includes ~700 new lines of daemon tests and 8 new test modules covering enrollment transactionality, crash/backoff, receipt reconciliation, credential-store permissions, origin pinning/redirect refusal, raw fingerprints, hook rollback and env isolation, and config masking)
- Frontend: 9 vitest component/API tests passed; `tsc --noEmit` clean; production build verified
- Schema advances `user_version` to 8 (v7 foundation + v8 recovery metadata), additive per convention

## Remaining external work

The hosted recurring-enrollment, submission, receipt-reconciliation, and revoke endpoints ship separately in the private broker; a staging canary is required before production rollout.

https://claude.ai/code/session_01PcJsydoENPknWvoFE2mWvX


kai-rayward:
## Deep review — recurring auto-upload (PR #115)

This is a thorough multi-agent review of the full diff (13 finders across the subsystems + security/concurrency/contract/test/docs lenses, each finding then run past 3 independent adversarial verifiers with a majority-vote gate). 38 raw → 30 deduped → **24 confirmed, 6 refuted**. I independently reproduced the highest-impact ones (noted inline). Line numbers are against this branch's head.

Nothing here is an egress-of-unredacted-content hole — the gate chain holds. The theme instead is the **automatic state machine getting durably stuck or storming the server**, plus a couple of consent/authority gaps and a total Windows breakage. Triage these; don't apply blindly.

### Major — automation gets stuck or hammers the server

- **Backoff is bypassed whenever a pending artifact exists** — `clawjournal/auto_upload.py:3553-3555` + `:2930`. *(Reproduced.)* After a retryable submit failure the share is left `sealed`/`submitting` and `next_retry_at` is stamped, but `_hook_due_check_on_connection` returns `due=True` on `recovery_pending`/`submitting_pending` **without reading `next_retry_at`**, and the runner's only backoff gate (`not force and pending_at_start is None and not decision.due`) is disabled by the pending row. Every Claude/Codex SessionStart then spawns a full cycle (capabilities fetch + submit POST) against a struggling server — and this also defeats the crash-backoff I added. The `submitting_pending` branch further ignores `mode`, so a disabled-but-unreconciled enrollment keeps spawning runners too.
- **`ControlChanged` records neither backoff nor action_required** — `clawjournal/auto_upload.py:3224`. A durable control-state mismatch (e.g. `source=all` but a non-claude/codex session gets indexed → `_current_scope` always returns `unsupported_source` → `_assert_control_state` raises `ControlChanged` every cycle) leaves `health='ready'`, `next_retry_at=NULL`, `last_completed_at` unadvanced — so `due_decision` stays due and every SessionStart relaunches two strict scans + two hosted calls forever, silently. (Same class as the finding above; the `except ControlChanged` handler should record backoff or action_required.)
- **Empty second-scan is misclassified as `payload_too_large` → durable `action_required`** — `clawjournal/auto_upload.py:3024`. If the top candidate resumes between the two strict scans (revision changes, `revision_stable_since` resets) the second report is empty, `_ranked_size_prefix` returns `([], 0)`, and the `if not selected` branch raises non-retryable `payload_too_large`. That converts to a durable `action_required` that blocks all future cycles with a false diagnosis — while the *first*-scan path returns a clean `nothing_new` for the identical condition.
- **`enrolled_at` boundary mismatch permanently breaks reauthorization** — `clawjournal/auto_upload.py:1095`. Create stores `enrolled_at = max(local intent, server enrolled_at)` (by design, tested), but the update path requires the server's returned `enrolled_at` to equal the stored boundary. If the local clock led the server at enroll time, they diverge forever: every scope/terms review or credential rotation raises `malformed_enrollment_response` → `paused`/`action_required`, unrecoverable without a full disable + re-enroll.
- **`splitlines()` hard-fails valid JSONL containing U+2028/U+2029/U+0085** — `clawjournal/parsing/parser.py:187`. *(Reproduced: `json.loads` accepts the line; strict parse raises `malformed JSONL input at line 1`.)* `JSON.stringify` (Claude Code) emits U+2028 unescaped inside strings; `str.splitlines()` splits on it, so a single valid record becomes two fragments. In a strict scan this marks the whole source `parse_failed` → every weekly cycle returns `strict_scan_incomplete` **forever** (the historical file never changes) while normal indexing of the same file succeeds. Split on `\n` only, matching the non-strict path.
- **`KeyError('backend')` on a zero-session share** — `clawjournal/workbench/daemon.py:1622` vs `:1758`. *(Confirmed structurally.)* The empty-sessions early return omits the new `"backend"` key, but `finalize_share_export_for_upload` reads `pii_summary["backend"]` unconditionally. A share whose sessions all get excluded (e.g. project added to `excluded_projects`, forcing a rebuild) → `KeyError` outside the try/except → manual endpoints 500, or `runner_crash` + backoff on the auto path.
- **Windows credential store is wholly inoperable** — `clawjournal/auto_upload_credentials.py:151`. `powershell.exe -Command <script> <path>` never populates `$args` (only `-File` does; `-Command` concatenates trailing tokens into the command text), so `$target` is `$null`, `Test-Path -LiteralPath $null` is a terminating error under `$ErrorActionPreference='Stop'`, and **every** credential write/load/disable/recovery raises `CredentialStoreError`. Fail-closed, but the feature can't enroll — or even revoke — on the one platform this ACL code targets, with no test (POSIX CI never runs the `nt` branch). Use `-File` with a script file, or pass the path via an env var / `-EncodedCommand`.
- **Interactive `enable` crashes with a raw traceback on ordinary input errors** — `clawjournal/cli_auto_upload.py:137`. The email-verification path calls `request_email_verification` / `confirm_pending_email_verification` / `input()` with no handler, so a non-academic email (`ValueError`), a mistyped code (`HostedServiceError`), a network blip (`URLError`), or Ctrl-D (`EOFError`) dumps a traceback and forces restarting the whole consent flow. The sibling `verify-email` command wraps the identical calls in `except (OSError, RuntimeError, ValueError)` and exits 1 cleanly — mirror that.

### Major — consent, authority, accessibility

- **Consent binds the terms versions but not the displayed scope** — `clawjournal/auto_upload.py:856`. Acceptance compares only `accepted_authorization_version`/`accepted_retention_version`; the accepting call re-runs a strict scan and recomputes scope. A project (C) that appears between challenge-display and acceptance is enrolled even though the consent screen listed only [A, B], and the per-cycle scope-drift check can't catch it because C is now part of the snapshot. Bind a scope hash from the challenge into acceptance.
- **`disable()` can be silently resurrected by an in-flight `enable()`** — `clawjournal/auto_upload.py:1442`. `disable()` takes no whole-run lock and skips the generation CAS bump when the row is `mode='off'` — exactly the intent state `enable()` persists during its network window. Concurrent disable during a fresh enroll: disable sees `off`, bumps nothing, uninstalls hooks, finds no credentials yet, reports "disabled"; then enable's final CAS commit (unchanged generation) lands `mode='enabled'` with live credentials on disk. The user was told it's off but weekly uploads can proceed — violates "controls win / disable removes authority first."
- **Authorization dialog has no focus management** — `clawjournal/web/frontend/src/components/AutoUploadControls.tsx:384`. `role=dialog`/`aria-modal` but no initial focus move, no Tab trap, no focus restore, and non-Escape keys aren't swallowed — so Tab reaches the still-enabled background Pause/Turn-off buttons and Enter activates them *underneath* the open consent modal (the overlay blocks clicks, not keys). The project's own `ConfirmDialog.tsx` already implements the full trap; this consent-critical modal regresses that contract.

### Minor — robustness, correctness, UX (full list; fix as a batch)

- **`daemon.py:1741`** — `finalize_share_export_for_upload`'s blanket `except Exception` swallows the `ControlChanged` that `pii.py`'s `_AgentCallGateError` plumbing carefully re-raises, so a pause during AI-PII review is recorded as retryable `package_failed` instead of taking the `ControlChanged` path. (No egress — controls still win — but telemetry/state misreport.)
- **`auto_upload_client.py:295`** — `_open_request` classifies only `TimeoutError`/`URLError`/`OSError`; `http.client.HTTPException` subclasses (`BadStatusLine`, `IncompleteRead`) escape unclassified, so `disable()` crashes unhandled instead of entering the revocation retry path, and `submit` skips the `ambiguous` marking.
- **`auto_upload_client.py:508`** — a 3xx to the submission POST yields `redirect_rejected` (`retryable=False, ambiguous=False`), so the runner marks the sealed artifact terminally `rejected` even though the body already crossed egress; every *other* post-egress-uncertain outcome is marked `ambiguous` and reconciles receipts first. A committed-server-side bundle would then have no local record.
- **`daemon.py:2039`** — `_final_manual_share_egress_gate` re-checks release/source/revision but not `share_predecessor_blockers`, though it's mutable during the minutes-long AI-PII pass; two revision-chain shares submitted back-to-back can upload a stale-predecessor duplicate that #109 added that check to refuse.
- **`auto_upload.py:1156`** — the non-rotating update path (healthy active token, PATCH) never sends `verified_email_token`, yet the success path unconditionally pops it from config, deleting a still-valid manual hosted-upload credential as a side effect of a recurring-scope review (the "consumed" comment is false on this path).
- **`auto_upload.py:3169`** — seal-boundary failures (`_assert_control_state` :3149, `_write_sealed_zip` :3166, `_seal_share_ledger` :3169) don't call `_cleanup_unsent_draft` like the neighboring checks do, orphaning a draft share row + export dir + a fully written `auto-upload.sealed.zip` that nothing sweeps (disable only removes `submission_state='sealed'` rows).
- **`raw_sources.py:79`** — directory fingerprint seeds `latest_mtime_ns` from the dir's own `st_mtime_ns`, contradicting the "unrelated directory files are ignored" docstring; touching any top-level entry produces a spurious `raw source changed` block even when every parser-input byte is identical (and it's internally inconsistent: mtime from the initial stat, dev/ino from the final stat).
- **`config.py:37` / `auto_upload.py:2308`** — `auto_upload_egress_lock` is held across raw-file SHA-256 re-hashing (Claude logs can be hundreds of MB) and an `open_index()`; meanwhile `save_config` blocks on it unbounded on POSIX (HTTP handler thread hangs) or drops the write after ~10s on Windows. The "short boundary lock" comment doesn't match the work inside it.
- **`daemon.py:143`** — `_strict_scan_log_guard` iterates `loggerDict.values()` without holding `logging._lock`; a concurrent first-time `getLogger()` in another daemon thread raises `dictionary changed size during iteration` → spurious `runner_crash` + backoff. Snapshot under the lock.
- **`AutoUploadControls.tsx:258`** — every dialog-dismiss path (Escape/backdrop/Cancel) is gated on `!loading`, and `loading` spans the whole challenge fetch (strict scan + forced hosted discovery, tens of seconds to minutes, browser fetch has no timeout) — the user is trapped in an unclosable modal on a purely read-only request.
- **`cli_auto_upload.py:101`** — server-controlled authorization/retention text and error messages are printed raw; a malicious/compromised hosted endpoint can embed ANSI control sequences to rewrite or hide the locally-computed Sources/Projects/Destination/cap consent lines before the user accepts. Strip C0/C1 controls (except `\n\t`) before printing.
- **`index.py:4772`** — `get_auto_upload_candidate_report` does a full `json.load` of every eligible session blob for the `missing_blob` gate, so the 2.5s-polled status/preview endpoint is O(eligible sessions × blob size) in disk I/O — precisely while the runner holds SQLite locks. Status was meant to be a cheap read.
- **`auto_upload.py:1683`** — no test exercises the bundle-size budget: the `deferred_by_size > 0` branch, the empty-prefix `payload_too_large` branch, and the final sealed-ZIP size gate (`:3160`) are all uncovered, and this code decides which traces leave the machine.

### Refuted / not defects

The "`enable()` doesn't enforce a prior manual receipt" claims (×3) are wrong — the offer/capability gating enforces it upstream — and a `last_result_count=None` claim was refuted on inspection. Two low-priority **doc-precision** notes are worth a glance: `PRIVACY.md:12` says the recurring hook "is inert unless … enrollment remains enabled," but per the backoff finding above it can still spawn a *recovery* runner (no content egress) when a submission is pending while paused/off — the wording becomes accurate once that's fixed; and `PRIVACY.md:81` may now *understate* the code (manual share appears to also fail-closed on rules-only AI-PII coverage), i.e. the code is stricter than documented — verify and align the text.

---
*Generated by a Claude Code multi-agent review workflow; the majors and the reproduced items are high-confidence, the minors are worth a triage pass rather than blind application.*

PR #117, opened 2026-07-18
Add one-click desktop launcher

richardopenrefinery:
## Summary

- add `clawjournal desktop install|status|refresh|uninstall|launch`
- create a one-click **ClawJournal** shortcut that starts or reuses the local workbench and requests a fresh scan
- render Day 0–10 facial-expression icons locally, from a big smile to a dramatic crying face
- refresh the icon with user-level Windows Task Scheduler, macOS LaunchAgent, and Linux systemd/autostart integrations
- add `--desktop-shortcut` / `-DesktopShortcut` installer options and user documentation
- migrate and remove the previously managed **ClawJournal Workbench** shortcut during upgrades
- keep console-mode child tools, including Scoop shims, hidden when launched by the no-console desktop process

## Why

ClawJournal currently requires users to run `clawjournal scan` and `clawjournal serve` from a terminal. This adds an optional one-click desktop entry while preserving the local-only behavior.

On Windows, launching directly from the home directory can also make a checkout at `~/clawjournal` shadow the installed Python package. The generated no-console bootstrap now starts from `~/.clawjournal/desktop` and captures early startup errors in `launcher.log`.

Because the desktop entry uses `pythonw.exe`, console-mode children such as a Scoop-installed TruffleHog shim would otherwise allocate and briefly flash their own command window. The bootstrap applies `CREATE_NO_WINDOW` to non-detached child processes, preserving scans and scoring without visible terminal windows.

## User impact

The shortcut opens the local browser workbench, scans in the background, and never uploads data. The last-opened timestamp and generated icons remain under `~/.clawjournal/desktop/`. Windows refresh tasks are allowed to run on battery and are capped at five minutes.

## Validation

- `python -m pytest tests/test_desktop.py tests/test_cli.py -q` — **168 passed**
- Windows integration smoke test:
  - built the production frontend
  - installed and invoked the generated `.lnk`
  - confirmed the private bootstrap and corrected working directory
  - confirmed the installed bootstrap applies Windows no-console flags to child tools
  - confirmed the workbench returned HTTP 200 at `localhost:8384`
  - ran the scheduled icon refresh successfully with result 0
- A broader Windows/Python 3.14 run completed with 2,445 passed, 2 skipped, and 72 existing platform-sensitive failures outside this change (encoding, path separator, symlink, and POSIX mode assumptions). The repository CI matrix targets Python 3.10–3.13.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,8 @@ $env:Path = [Environment]::GetEnvironmentVariable('Path','Machine') + ';' + [Env
 git clone https://github.com/rayward-external/clawjournal.git ~/clawjournal
 cd ~/clawjournal
 ./scripts/install.sh --with-frontend       # or: sh scripts/install.sh --with-frontend
+# Optional: build the workbench and add a one-click desktop shortcut:
+./scripts/install.sh --desktop-shortcut
 ```
 
 **Install — native Windows PowerShell** (use `powershell` if `pwsh` isn't installed):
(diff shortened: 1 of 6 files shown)


richardopenrefinery:
Randomly terminal or a blank command prompt is randomly popping up due to Scoop.
Trying to fix it.

PR #118, opened 2026-07-18
Add one-click ClawJournal desktop launcher

richardopenrefinery:
## Summary

- add `clawjournal desktop install|status|refresh|uninstall|launch`
- create a one-click **ClawJournal** shortcut that starts or reuses the local workbench and requests a fresh scan
- render Day 0–10 facial-expression icons locally, from a big smile to a dramatic crying face
- refresh the icon with user-level Windows Task Scheduler, macOS LaunchAgent, and Linux systemd/autostart integrations
- add `--desktop-shortcut` / `-DesktopShortcut` installer options and user documentation
- migrate and remove the previously managed **ClawJournal Workbench** shortcut during upgrades
- keep console-mode child tools, including Scoop shims, hidden when launched by the no-console desktop process
- make the desktop tests timezone- and path-platform-independent
- move desktop-open state changes behind an authenticated POST and schedule OS icon work off the HTTP thread
- authenticate an existing daemon with a nonce-based HMAC proof before sending it the bearer token
- reject foreign services on the configured port and recheck daemon identity after bind races
- serialize scanner passes and coalesce repeated desktop scan requests while a scan is active
- cap desktop platform commands at 30 seconds so icon integration cannot hang indefinitely

Supersedes #117 with the CI portability corrections included.

## Why

ClawJournal currently requires users to run `clawjournal scan` and `clawjournal serve` from a terminal. This adds an optional one-click desktop entry while preserving the local-only behavior.

The workbench shell is intentionally unauthenticated, so serving `index.html` must not mutate desktop state or run OS commands. A successful SPA mount now sends an authenticated notification instead, and the daemon handles the icon refresh asynchronously.

The launcher also verifies that the configured port belongs to this ClawJournal installation before opening it or sending an authenticated scan request. The verification uses a fresh public challenge and an HMAC response, keeping the bearer token away from unknown listeners.

On Windows, launching directly from the home directory can make a checkout at `~/clawjournal` shadow the installed Python package. The generated no-console bootstrap now starts from `~/.clawjournal/desktop` and captures early startup errors in `launcher.log`.

Because the desktop entry uses `pythonw.exe`, console-mode children such as a Scoop-installed TruffleHog shim would otherwise allocate and briefly flash their own command window. The bootstrap applies `CREATE_NO_WINDOW` to non-detached child processes, preserving scans and scoring without visible terminal windows.

## User impact

The shortcut opens the local browser workbench, scans in the background, and never uploads data. The last-opened timestamp and generated icons remain under `~/.clawjournal/desktop/`. Repeated clicks reuse only a verified ClawJournal daemon and coalesce scans rather than running them concurrently. Windows refresh tasks are allowed to run on battery and are capped at five minutes.

## Validation

- `PYTHONPATH=$PWD pytest -q` — **2,977 passed, 4 skipped**
- focused desktop, scan, API authentication, and static-serving regressions — **41 passed**
- frontend Vitest suite — **40 passed**
- production frontend build — passed
- `python -m compileall -q clawjournal` — passed
- `sh -n scripts/install.sh` — passed
- `git diff --check` — clean
- Windows integration smoke test:
  - built the production frontend
  - installed and invoked the generated `.lnk`
  - confirmed the private bootstrap and corrected working directory
  - confirmed the installed bootstrap applies Windows no-console flags to child tools
  - confirmed the workbench returned HTTP 200 at `localhost:8384`
  - ran the scheduled icon refresh successfully with result 0

The branch includes the current `main` before these fixes.


kai-rayward:
Reviewed and pushed fixes for the issues found (97a8bfb). Full suite green: 2528 passed.

**Windows icon refresh could stall silently.** The daily task ran at a fixed 09:00 with no `-StartWhenAvailable`, so a machine asleep at that hour skipped the day entirely. Linux has a login autostart and `Persistent=true`, macOS has `RunAtLoad` — Windows had only the fixed-time task plus refresh-on-open, and "open the app" is exactly the behavior the sad face exists to prompt. Added the flag.

**`launcher.log` grew without bound.** The shortcut redirects the daemon's stdout/stderr there, and `run_server` logs every HTTP request. Now capped at 1 MB, trimmed to the most recent half on launch. Trimming in place rather than rotating is deliberate — the POSIX launchers hand the daemon an O_APPEND descriptor tied to the open inode, so a rename would leave it writing to the orphan.

**`note_opened()` ran in hot-reload children.** It was called before the `RELOAD_CHILD_ENV` check, so every dev-mode reload rewrote `last_opened` and re-ran a refresh (a PowerShell invocation per reload on Windows). Moved below the check.

**A pinned tab never reset the clock.** With a long-lived daemon, only shortcut launches and `serve` starts counted as opens, so someone using the workbench daily could still watch the icon descend into dramatic crying. Serving `index.html` now counts. To make that safe on the page-serving path, `note_opened()` is debounced to the calendar day — the icon only changes on a day boundary, so a same-day open needs no shortcut rewrite.

**Ownership was provable from state rather than from the artifact.** `_ensure_available` returned early whenever `install.json` existed, skipping the ownership check entirely — so if a user deleted the managed shortcut and later put their own file at that path, reinstall clobbered it. `_install_macos` had the same hole. All three platforms now prove ownership from the artifact itself (desktop-entry marker, `CFBundleIdentifier`, and a new `.lnk` byte-marker scan, since `.lnk` has no comment field but stores target/argument strings verbatim in both ASCII and UTF-16LE). Applied the same check to the removal paths, which previously deleted Windows shortcuts unconditionally.

**Redundant work on Windows install.** `install()` ended with `refresh(quiet=True)`, rewriting the `.lnk` through a second PowerShell run right after the installer had written it; `_write_windows_bootstrap()` was also called twice. Dropped both, moving the shell notification into `_install_windows`.

Added tests for the debounce, the clobber refusal, foreign-file uninstall, the unknown-platform error, `.lnk` ownership in both encodings, log trimming, and naive stored timestamps.

One thing left for you rather than me: the escalating guilt-face is an engagement mechanic, and it adds three OS persistence surfaces (Task Scheduler, LaunchAgent, systemd + autostart) to a tool whose pitch is local-first and quiet. It's opt-in and fully local, so nothing here is a privacy concern — but if you want the feature with less machinery, dropping the daily task and refreshing only at login/open removes most of the moving parts at the cost of same-day icon accuracy. Your call.

kai-rayward:
Pushed the isolation fix and both HIGH issues (dbabc7a). Suite green: 2533 passed.

**Desktop state escaped test isolation.** `desktop.py` did `from .config import CONFIG_DIR` and built `STATE_DIR`/`STATE_FILE`/`LAST_OPENED_FILE`/`ICONS_DIR`/`LOG_FILE`/`WINDOWS_BOOTSTRAP` as module constants. The autouse `tmp_config` fixture patches `clawjournal.config.CONFIG_DIR`, which cannot rebind constants already captured at import — verified directly:

```
desktop.STATE_FILE at import: /Users/<me>/.clawjournal/desktop/install.json
after patching clawjournal.config.CONFIG_DIR -> /tmp/fake-isolated-config
desktop.STATE_FILE is STILL:  /Users/<me>/.clawjournal/desktop/install.json
```

That mattered because the `note_opened()` hook I added to `_serve_static` is reached by `tests/workbench/test_daemon.py`, which points `FRONTEND_DIST` at a real `index.html`. On a machine that had run `desktop install`, `pytest` would read the real `install.json`, write the real `last_opened`, and rewrite the real desktop shortcut (spawning PowerShell on Windows). Same class of failure as the config clobbering that made the egress gates look like they reset on their own. All six paths now resolve per call, so the existing autouse fixture covers this module — no conftest change needed, and the next module to touch `CONFIG_DIR` doesn't inherit the trap.

**`refresh()` never checked ownership, and it runs unattended.** It's driven by schtasks, the LaunchAgent, the systemd timer, and now the daemon's page handler. `_refresh_linux` and `_refresh_windows` wrote over whatever was at the shortcut path, and `_refresh_windows` didn't check existence at all — so deleting the Windows icon got it recreated at 09:00 the next morning, and every morning after. An icon you can't delete reads as malware. All three refreshers now go through `_require_managed()`: present, and still ours, or `DesktopError`. `note_opened()` already swallows that, so page loads are unaffected.

**A partial macOS install bricked install *and* uninstall.** The bundle was built in place with `Info.plist` — its only ownership marker — written last. Any failure in between left a `.app` that `_macos_app_is_managed` disowned permanently: `install()` raised "Refusing to replace" forever, and `uninstall()` returned `removed=False` because state was never written, with nothing telling the user a manual `rm -rf` was the only way out. The bundle is now staged beside the target and swapped in only once complete, with the staging dir cleaned up on failure.

Four new tests, each confirmed to **fail against the pre-fix module and pass against the fixed one** — including a direct test of the Windows refresh guard, since the Linux path already raised on a missing shortcut and would have masked that bug.

Still open from the review, not addressed here:
- **Double-click spawns a second daemon.** Reproduced: `_workbench_running()` uses a 0.6s auth'd `/api/stats` call (a real SQL aggregate) as a liveness probe; when it times out, `run_server` fails to bind and silently falls back to an ephemeral port, giving you two daemons and two scanners on one SQLite index. `cli.py:1795` already has `_daemon_port_is_open()` — a 0.25s TCP connect — which is the right primitive. Needs a call on whether `launch()` reuses it or `run_server` stops falling back.
- **DST off-by-one.** Stored `2026-11-01T00:30-04:00`, now `2026-11-01T20:00-05:00` — same local date — yields day 1 and `opened_today() == False`. Caused by reprojecting a past instant through today's fixed offset; comparing `opened.date()` to `current.date()` directly fixes it.
- **Every icon has a dark halo.** `_blend` emits premultiplied RGB into a straight-alpha (color type 6) PNG. Measured: the Day-0 shadow is `(17,11,4,55)` where straight alpha needs `(79,53,18,55)`, so the drop shadow is near-black instead of warm brown and every antialiased edge is dark-fringed.
- `_trim_log` empties the log when the trimmed tail has no newline (3 MB single-line log to 0 bytes) — my regression, plus a non-atomic read-then-truncate window.
- Minor: `_remove_previous_shortcut` can't clean up a relocated desktop dir (OneDrive/xdg); `_macos_app_is_managed` doesn't catch `ExpatError`; the Windows bootstrap discards `main()`'s return value; `_ico_from_png` hardcodes 256x256; `_desktop_exec_quote` under-escapes backslashes.

Coverage context for whoever picks these up: `desktop.py` is at 63%, and the shape is the issue — the fixture pins `_platform` to `linux`, so the Windows and macOS install paths never execute, `run_desktop_command` is at 1/26 statements with no test invoking the subcommand, and `_workbench_running`/`_request_scan` are fully stubbed. That last one is exactly why the double-daemon bug went unnoticed.

kai-rayward:
Pushed the double-daemon fix (1d357ef). Suite green: 2536 passed.

**The probe.** `_workbench_running` now delegates to `cli._daemon_port_is_open` — the 0.25s TCP connect already written for this exact question — instead of calling `/api/stats` with a 0.6s timeout. Liveness here only ever needed to answer "is the port taken"; asking a real SQL aggregate over the index meant a daemon that was merely busy scanning read as absent. Measured against a deliberately slow daemon on 8384:

```
before:  _workbench_running(8384) -> False   (0.6s /api/stats timeout)
after:   _workbench_running(8384) -> True    (in 0.030s)
```

**The race.** That still left the narrow window where two clicks both find the port free and both try to bind — and the loser was landing on an ephemeral port silently. `run_server` grows `allow_port_fallback`, defaulting to `True` so `clawjournal serve` behaves exactly as before; the launcher passes `False` and joins the winner on `OSError` rather than starting a hidden duplicate. Same repro, after:

```
run_server: refused to take a random port (Address already in use) -> caller reuses 8384
```

Three new tests, all confirmed to fail against the pre-fix modules. The busy-daemon one drives a real unaccepting socket rather than a stub, so it exercises the probe itself rather than a mock of it — which matters here, since `_workbench_running` being fully stubbed is exactly why this went unnoticed. (Worth noting for anyone writing similar tests: the listener needs a backlog larger than the number of probes, since each unaccepted connect holds a slot — a backlog of 1 makes the second probe fail and looks like a code bug.)

One deliberate behaviour change to flag: if some unrelated service occupies 8384, the launcher now opens the browser at it instead of silently starting the workbench on a random port. I think that is the better failure — visible rather than hidden, and no stray daemon — but it is a judgement call, so say the word if you would rather it detect the foreign service and report instead.

Remaining from the review, all lower severity:
- **DST off-by-one** — stored `2026-11-01T00:30-04:00`, now `2026-11-01T20:00-05:00` (same local date) yields day 1 and `opened_today() == False`. Reprojecting a past instant through today's fixed offset; compare `opened.date()` to `current.date()` directly.
- **Dark halo on every icon** — `_blend` emits premultiplied RGB into a straight-alpha (type 6) PNG; the Day-0 shadow measures `(17,11,4,55)` where straight alpha needs `(79,53,18,55)`.
- **`_trim_log`** empties the log when the trimmed tail has no newline, plus a non-atomic read-then-truncate window.
- Minor: `_remove_previous_shortcut` can't clean up a relocated desktop dir; `_macos_app_is_managed` misses `ExpatError`; the Windows bootstrap discards `main()`'s exit code; `_ico_from_png` hardcodes 256x256; `_desktop_exec_quote` under-escapes backslashes.

kai-rayward:
Pushed the remaining items (cfb5ede). Suite green: 2544 passed. That clears everything from the review.

**DST off-by-one.** `days_since_last_opened` and `opened_today` reprojected the stored instant through *today's* UTC offset, which shifts the date across a fall-back boundary — an open at `2026-11-01T00:30-04:00` read as `2026-10-31` once the clocks went back, so the icon turned sad the same afternoon it was used. The stamp already records the local date it was written, so both now compare wall-clock dates directly.

**Premultiplied alpha.** Worth spelling out, because `_blend` is not actually wrong: it composites in premultiplied space, which is exactly what makes the supersample box filter correct. The bug was only that `png()` emitted those values into a colour-type-6 PNG, which stores *straight* alpha. So the fix is one un-premultiply pass after downscaling, not a rewrite of the compositor. Measured on the Day-0 shadow:

```
before:  (17, 11, 4, 55)      # premultiplied -> near-black
after:   (79, 51, 19, 55)     # straight alpha, want (79, 53, 18, 55)
```

Strongly-covered edge pixels now sit at `(174, 111, 15)` against a rim colour of `(181, 116, 15)`; premultiplied fringing put them near `(145, 93, 12)`. The residual drift is 8-bit quantization at low alpha, which is inherent to premultiplied storage and not worth chasing.

**`_trim_log` emptied a newline-free log** — a single enormous line meant no newline in the kept window, so seeking past the first boundary discarded everything (3 MB to 0 bytes). Falls back to the raw tail, and uses `r+b` so the file is never momentarily zero-length.

**`_remove_previous_shortcut` required the same parent directory**, so it only ever caught a rename. The case that actually happens is a *move* — OneDrive redirection, or `xdg-user-dir DESKTOP` changing — which orphaned the old icon permanently, since `install.json` is rewritten to the new path immediately after. Ownership is still proven from the artifact before anything is deleted.

Also fixed: `_macos_app_is_managed` missed `ExpatError` (plistlib raises that, not `InvalidFileException`, for malformed XML) and crashed `install()` with a traceback on a foreign bundle; the generated Windows bootstrap logged a full traceback for an ordinary `SystemExit`; `_ico_from_png` hardcoded the 256 sentinel; `_desktop_exec_quote` under-escaped backslashes, which pass through two unescaping layers and need four.

Eight new tests, each confirmed to fail against the pre-fix module. The alpha one decodes the generated PNG rather than inspecting the canvas, so it pins the actual file contents.

Across the four fix commits the test file went from 8 tests to 31. The coverage shape is still worth knowing for future work: the fixture pins `_platform` to `linux`, so the Windows and macOS install paths are only exercised where a test overrides it, and `run_desktop_command` still has no test invoking the subcommand end to end.

kai-rayward:
## Stepping back: is this worth merging, and does it help sharing?

After four fix rounds the code is in good shape — 2544 passing, thirteen defects fixed with regression tests. Two higher-level points for the merge decision, separate from code quality.

### Does it make data sharing easier?

I checked this against the actual share funnel rather than by impression: **the launcher touches zero share-path code.** No gate (`--source`, `--confirm-projects`, `verified_email`), no hold-state, no redaction, no export, no findings. It starts-or-reuses the daemon, calls `/api/scan`, and opens the workbench. It's a convenience wrapper around `scan` + `serve`, sitting at the very front of the six-stage flow.

Everything that stands between a user and a share lives *downstream* of where this PR stops: confirming source + project scope, triage/scoring, the hold→release state machine, two-pass redaction + the mandatory TruffleHog gate, and the separate opt-in upload. The launcher deposits you at the workbench door; every one of those stages is exactly as many clicks as before. So for "more people complete a share," this moves a step that wasn't the bottleneck.

The mood icon is a re-engagement mechanic — it nags you to *open* the app. But opening was never the hard part; the friction is the redaction/hold/confirm gates, which exist on purpose in a privacy-first tool. A guilt-face that drives daily opens doesn't move anyone across those gates, it just raises how often they bounce off them — and it pays for that with three OS persistence surfaces (Task Scheduler, LaunchAgent, systemd + autostart) on a tool that markets itself as local-first and quiet.

### Merge recommendation

The PR is really two features welded together, with very different cost profiles (measured on the current head):

| Part | Size | Character |
|---|---|---|
| One-click launcher | ~640 lines | start/reuse daemon, scan, open workbench; bugs were ordinary bugs |
| Mood-icon machinery | ~330 lines / 19 fns | hand-rolled rasterizer + 3 binary formats + 16 scheduled-task/persistence call sites; produced the gnarliest defects (DST, premultiplied alpha, the unattended-refresh hole that made a Windows icon undeletable) |

The launcher is straightforwardly useful and I'd merge it without hesitation. The mood icon is orthogonal to sharing, in mild tension with the product's stance, and is where the maintenance risk and the un-CI-able surface concentrate — at 63% coverage with `_platform` pinned to `linux`, neither the Windows nor macOS install path runs under test, so its remaining risk needs a manual smoke test on real machines, not another review pass.

**Suggestion:** split the icon out and merge the launcher alone. If you do want the guilt-face, merge as-is but smoke-test install/uninstall on real Windows and macOS before relying on it. Either way, more review iterations are past the point of diminishing returns — the code reads clean; what's left is a product call about whether the icon earns its cost.

PR #120, opened 2026-07-19
fix(auto-upload): harden Windows credentials and gate rollout UI

JustMeStudio:
## Summary

This PR prepares recurring automatic uploads for a limited rollout.

It:

- gates the Automatic Upload settings UI behind `CLAWJOURNAL_ENABLE_AUTO_UPLOAD_UI=1`
- keeps the CLI and local API available for internal testing
- keeps the controls visible for already-enrolled users so they can pause or revoke access
- fixes Windows credential-directory ACL handling
- preserves the existing capped, future-only recurring-upload behavior

## Rollout behavior

For a new user, the Automatic Upload UI is hidden by default.

Internal testers can enable it before starting the workbench:

```powershell
$env:CLAWJOURNAL_ENABLE_AUTO_UPLOAD_UI="1"
clawjournal serve
```

Removing the variable hides the UI again for users who are not enrolled:

```powershell
Remove-Item Env:CLAWJOURNAL_ENABLE_AUTO_UPLOAD_UI -ErrorAction SilentlyContinue
clawjournal serve
```

Users with an active, paused, or revocation-pending enrollment continue to see the controls even without the rollout flag. This ensures that hiding the enrollment entry point never prevents an existing participant from pausing or revoking authorization.

## Local end-to-end validation

I tested the complete client-to-server recurring-upload workflow against a locally deployed `clawjournal-share` service backed by PostgreSQL and filesystem bundle storage.

### Environment

- Windows PowerShell
- Python virtual environment with the client installed in editable mode
- Codex CLI
- Docker Compose
- Local `clawjournal-share` application
- Local PostgreSQL
- Local bundle storage mounted at `/var/clawjournal/bundles`
- Hosted-share origin: `http://localhost:8080`
- Workbench origin: `http://localhost:8384`

The local-only recurring-upload HTTP exception was explicitly enabled:

```powershell
$env:CLAWJOURNAL_SHARE_URL="http://localhost:8080/share"
$env:CLAWJOURNAL_ALLOW_INSECURE_LOOPBACK_RECURRING="1"
$env:CLAWJOURNAL_ENABLE_AUTO_UPLOAD_UI="1"
$env:NO_PROXY="localhost,127.0.0.1,::1"
```

Production HTTPS validation remains unchanged.

### Server startup

I started a clean local service stack with:

```powershell
docker compose -p clawjournal-v1test up --build -d
docker compose -p clawjournal-v1test ps
```

I verified the hosted capability document:

```powershell
curl.exe --noproxy "*" \
  http://localhost:8080/.well-known/clawjournal-share.json
```

The service reported:

- `recurring_upload_api_version: 1`
- recurring enrollment open
- recurring submissions open
- maximum five recurring sessions
- exact artifact idempotency enabled
- duplicate revision enforcement enabled

### Manual-share prerequisite

I completed a normal hosted manual share through the existing Share workflow.

The server accepted the bundle and returned a hosted receipt. This established the manual-share prerequisite before recurring enrollment was offered.

### Enrollment flow

From Settings, I:

1. opened **Review and enable**
2. selected Codex as the agent source
3. reviewed the exact source and project snapshot
4. reviewed the recurring authorization and retention text
5. completed email verification
6. accepted the authorization checkbox
7. enabled automatic uploads

The resulting client state reported:

```json
{
  "mode": "enabled",
  "health": "ready",
  "overlay": null,
  "authorization": {
    "version": "consent-v1"
  },
  "retention": {
    "version": "retention-v1"
  },
  "scope": {
    "sources": ["codex"]
  }
}
```

The UI displayed:

- Automatic Uploads: On
- Codex hook: configured
- a future-only enrollment cutoff
- a seven-day cadence
- a five-trace per-cycle cap
- the next scheduled due time

### Future-only session test

After enrollment, I created a new persisted Codex session:

```powershell
codex exec \
  --model gpt-5.4-mini \
  -c 'model_reasoning_effort="low"' \
  "Inspect README.md and return one short improvement suggestion."
```

I did not use `--ephemeral`, so the session was saved and available to ClawJournal.

A refreshed candidate preview correctly separated:

- sessions completed before enrollment
- sessions outside the enrolled source
- newly completed append-only sessions that had not reached the 24-hour stability period

### Immediate local cycle test

The production behavior requires an append-only Codex trace to remain unchanged for 24 hours.

To test the rest of the workflow without waiting 24 hours, I changed `AUTO_UPLOAD_STABILITY_HOURS` to zero only inside a temporary Python process:

```powershell
@'
import json
from clawjournal.workbench import index

index.AUTO_UPLOAD_STABILITY_HOURS = 0

from clawjournal.auto_upload import run_cycle

result = run_cycle(force=True)
print(json.dumps(result, ensure_ascii=False, indent=2))
'@ | python -
```

This bypassed only the local stability delay. Candidate selection, redaction, secret scanning, artifact construction, authorization checks, upload, persistence, and receipt handling still used the real implementation.

The cycle returned:

```json
{
  "ok": true,
  "code": "uploaded",
  "count": 1,
  "receipt_reference": "cj_ly4PvU2h_JuMEv3-VKxXXWOHqNvCuJNK",
  "client_submission_id": "637d397e-4dfd-4744-ace0-47784b7b9ae3",
  "artifact_sha256": "a698804372e73f57fe2b8c8011989369683f771734731c1c0ef51d4ab6baa5f5",
  "deferred_by_size": 0
}
```

### Client-state verification

After upload, `clawjournal auto-upload status --json` reported:

- `mode: enabled`
- `health: ready`
- `overlay: null`
- `last_result.code: uploaded`
- `last_result.count: 1`
- the same receipt reference returned by the server
- a new `next_due_at` seven days later

This confirms the successful cycle advanced the recurring schedule.

### PostgreSQL verification

I queried the recurring submission table:

```sql
SELECT
    client_submission_id,
    receipt_id,
    status,
    artifact_hash,
    object_path,
    created_at
FROM recurring_upload_submissions
ORDER BY created_at DESC
LIMIT 5;
```

The server stored:

- the expected client submission ID
- the expected receipt ID
- status `received`
- the expected artifact SHA-256
- a ZIP path under `/var/clawjournal/bundles`
- the upload timestamp

The stored artifact hash exactly matched the client result:

```text
a698804372e73f57fe2b8c8011989369683f771734731c1c0ef51d4ab6baa5f5
```

### Receipt verification

I queried the hosted receipt endpoint:

```powershell
Invoke-RestMethod \
  http://localhost:8080/api/receipts/cj_ly4PvU2h_JuMEv3-VKxXXWOHqNvCuJNK |
  ConvertTo-Json -Depth 5
```

The receipt reported:

- status `accepted`
- one accepted session
- `consent-v1`
- `retention-v1`
- the same ZIP hash as the client and PostgreSQL
- nine redactions
- no deduplication on the first submission

Redactions included:

- private URLs
- project names
- username data

### Storage verification

I verified that the ZIP existed in the server container:

```powershell
docker compose -p clawjournal-v1test exec app sh -lc \
  "find /var/clawjournal/bundles -type f -name '*.zip' -ls"
```

The uploaded artifact was stored at:

```text
/var/clawjournal/bundles/2026-07-19/cj_ly4PvU2h_JuMEv3-VKxXXWOHqNvCuJNK.zip
```

## Windows ACL validation

I also exercised recurring credential persistence on Windows.

The credential store successfully created and accessed:

```text
%USERPROFILE%\.clawjournal\credentials\auto_upload.json
```

The updated ACL handling allows the current Windows user to store recurring credentials while continuing to reject credential directories that cannot be secured to the current user.

This fixes the previous failure:

```text
could not establish a current-user-only Windows ACL
```

## UI-gating validation

I tested the hidden rollout behavior with a clean temporary user profile.

Without the rollout variable:

```powershell
Remove-Item Env:CLAWJOURNAL_ENABLE_AUTO_UPLOAD_UI -ErrorAction SilentlyContinue
clawjournal auto-upload status --json
```

The client returned:

```json
{
  "mode": "off",
  "ui_visible": false
}
```

The Automatic Upload panel was absent from Settings.

With the rollout variable enabled, the panel and enrollment workflow became available.

I also confirmed that an existing enabled enrollment remains visible after removing the rollout variable, preserving access to Preview, Run now, Pause, Review scope and terms, and Turn off.

## Automated validation

The focused client test suites cover:

- recurring enrollment state transitions
- Windows credential storage and ACL validation
- automatic-upload UI visibility rules
- visibility for existing enrollments
- hidden state for new users without the rollout flag
- authorization and retention challenge handling
- pause, resume, disable, and revocation states
- candidate selection and capped cycles
- submission and receipt persistence

## Remaining scheduling note

The real candidate-selection, artifact, upload, database, storage, and receipt path was tested end to end.

The seven-day wait was not observed in real time. For the end-to-end test, only the 24-hour append-only stability delay was overridden in-process. The successful cycle correctly persisted the next due time seven days later.

The Codex SessionStart hook was installed and reported as configured. The upload cycle itself was validated directly with `run_cycle(force=True)`; natural execution after the seven-day due time remains a scheduling observation rather than a blocking functional gap.

## Additional bug fixes

- Fixed Windows credential-store ACL operations hanging or timing out because each credential read/write launched multiple PowerShell processes. ACL setup now uses native Windows security APIs while preserving current-user-only access and fail-closed verification.
- Fixed recurring-upload authorization accepting stale responses after the dialog was closed and reopened, which could display outdated terms or interfere with a newer enrollment request.
- Fixed explicit collaborator emails being rejected client-side when they did not match an academic domain. When the hosted service advertises collaborator support, final allowlist validation is now delegated to the server.
- Fixed malformed destination URLs, invalid ports, credential-bearing URLs, and malformed IPv6 addresses escaping validation or causing raw exceptions later in the upload flow.
- Fixed session IDs containing Windows-invalid filename characters, reserved device names, path separators, or excessive lengths preventing indexing and automatic-upload eligibility. Unsafe IDs now use deterministic SHA-256 blob filenames.
- Fixed path-traversal risk from using raw session IDs as blob filenames while retaining compatibility with existing safe blob paths.
- Fixed Codex and OpenClaw session keys changing between Windows and POSIX because source paths were normalized using platform-specific `Path` behavior.
- Fixed scoring rubric and session blob reads relying on the Windows GBK default encoding instead of explicit UTF-8.

## Additional validation

- `146 passed, 1 skipped` — recurring-upload client, credentials, hooks, CLI, runner, and persistence tests.
- `243 passed` — workbench index and daemon tests.
- `70 passed` — scoring tests.
- `13 passed` — recurring-upload frontend tests.
- Production frontend TypeScript/Vite build completed successfully.
- `git diff --check` completed without errors.

### DIFF
--- a/clawjournal/auto_upload.py
+++ b/clawjournal/auto_upload.py
@@ -87,6 +87,7 @@ TELEMETRY_MAX_BYTES = 512 * 1024
 TELEMETRY_BACKUPS = 2
 SUPPORTED_HOOK_TARGETS = ("claude", "codex")
 HOOK_DB_BUSY_TIMEOUT_MS = 0
+AUTO_UPLOAD_UI_ENV = "CLAWJOURNAL_ENABLE_AUTO_UPLOAD_UI"
 _CONTROL_THREAD_LOCK = threading.Lock()
 
 # V1 supports only sources with both a SessionStart trigger and an audited,
@@ -557,6 +558,25 @@ def _has_successful_manual_receipt(conn: sqlite3.Connection) -> bool:
     )
 
 
+def _auto_upload_ui_visible(
+    config: Mapping[str, Any],
+    enrollment: Mapping[str, Any] | None = None,
+) -> bool:
+    explicitly_enabled = (
+        os.environ.get(AUTO_UPLOAD_UI_ENV) == "1"
+        or config.get("auto_upload_ui_enabled") is True
+    )
+    existing_authority = bool(
+        enrollment
+        and (
+            enrollment.get("mode") != "off"
+            or enrollment.get("server_enrollment_id")
+            or enrollment.get("revocation_pending")
+        )
+    )
+    return explicitly_enabled or existing_authority
+
+
 def _off_status(config: Mapping[str, Any]) -> dict[str, Any]:
     hooks = []
     for target in SUPPORTED_HOOK_TARGETS:
@@ -570,6 +590,7 @@ def _off_status(config: Mapping[str, Any]) -> dict[str, Any]:
         "run_now_allowed": False,
         "overlay": None,
         "pending_submission_state": None,
+        "ui_visible": _auto_upload_ui_visible(config),
         "offer_available": False,
         "scope": {"sources": [], "projects": []},
         "cap": MAX_SESSIONS,
@@ -671,6 +692,7 @@ def status(*, conn: sqlite3.Connection | None = None) -> dict[str, Any]:
             ),
             "overlay": overlay,
             "pending_submission_state": pending_submission_state,
+            "ui_visible": _auto_upload_ui_visible(config, enrollment),
             "offer_available": bool(
                 mode == "off"
                 and successful_manual
--- a/clawjournal/auto_upload_client.py
+++ b/clawjournal/auto_upload_client.py
@@ -72,7 +72,14 @@ def _normalized_origin(
     allow_local_http: bool = False,
     require_exact_origin: bool = False,
 ) -> str:
-    parsed = urlsplit(value)
+    try:
+        parsed = urlsplit(value)
+        parsed_port = parsed.port
+    except ValueError as exc:
+        raise CapabilityError(
+            "invalid_destination",
+            "Recurring upload requires an exact HTTPS destination origin.",
+        ) from exc
     hostname = (parsed.hostname or "").lower()
     scheme = parsed.scheme.lower()
     local = hostname in {"localhost", "127.0.0.1", "::1"}
@@ -94,7 +101,7 @@ def _normalized_origin(
             "invalid_destination",
             "Recurring upload requires an exact HTTPS destination origin.",
         )
-    port = f":{parsed.port}" if parsed.port is not None else ""
+    port = f":{parsed_port}" if parsed_port is not None else ""
     rendered_host = f"[{hostname}]" if ":" in hostname else hostname
     return f"{scheme}://{rendered_host}{port}"
 
--- a/clawjournal/auto_upload_credentials.py
+++ b/clawjournal/auto_upload_credentials.py
@@ -133,41 +133,84 @@ $target = $env:CLAWJOURNAL_ACL_TARGET
 $identity = [System.Security.Principal.WindowsIdentity]::GetCurrent().User
 $isDirectory = Test-Path -LiteralPath $target -PathType Container
 if ($isDirectory) {
-  $acl = New-Object System.Security.AccessControl.DirectorySecurity
-  $inheritance = [System.Security.AccessControl.InheritanceFlags]'ContainerInherit, ObjectInherit'
+  $grant = "*$($identity.Value):(OI)(CI)F"
 } else {
-  $acl = New-Object System.Security.AccessControl.FileSecurity
-  $inheritance = [System.Security.AccessControl.InheritanceFlags]::None
+  $grant = "*$($identity.Value):F"
 }
-$acl.SetAccessRuleProtection($true, $false)
-$acl.SetOwner($identity)
-$rule = New-Object System.Security.AccessControl.FileSystemAccessRule(
-  $identity,
-  [System.Security.AccessControl.FileSystemRights]::FullControl,
-  $inheritance,
-  [System.Security.AccessControl.PropagationFlags]::None,
-  [System.Security.AccessControl.AccessControlType]::Allow
+& icacls.exe $target /inheritance:r | Out-Null
+if ($LASTEXITCODE -ne 0) {
+  throw "icacls failed with exit code $LASTEXITCODE"
+}
+$existingAcl = Get-Acl -LiteralPath $target
+$otherIdentities = @(
+  $existingAcl.GetAccessRules(
+    $true,
+    $true,
+    [System.Security.Principal.SecurityIdentifier]
+  ) |
+    Where-Object { $_.IdentityReference -ne $identity } |
+    ForEach-Object { $_.IdentityReference.Value } |
+    Sort-Object -Unique
 )
-$acl.AddAccessRule($rule)
-Set-Acl -LiteralPath $target -AclObject $acl
+foreach ($otherIdentity in $otherIdentities) {
+  & icacls.exe $target /remove "*$otherIdentity" | Out-Null
+  if ($LASTEXITCODE -ne 0) {
+    throw "icacls could not remove an existing credential ACL entry"
+  }
+}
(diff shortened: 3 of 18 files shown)


kai-rayward:
## Review

**Verdict: one fix away from mergeable.** CI is green across the 3.10–3.13 matrix + smoke. The backend gating logic, the `icacls` rewrite, and the improved error surfacing all look good. One real frontend bug needs fixing first.

### Blocking: unreachable loading/error states in `AutoUploadControls.tsx`

The new early return in `AutoUploadPanel` sits above the existing loading and error branches:

```tsx
if (!status || !status.ui_visible) return null;   // new

if (!status && !loadError) { /* loading placeholder */ }  // now unreachable
if (!status) { /* error + Retry button */ }               // now unreachable
```

Because the new line returns on every `!status`, the loading placeholder and the error-with-Retry panel can never render. Concretely: an **enrolled** user whose status fetch fails now sees nothing at all — no error, no Retry, no Pause/Turn off controls. That contradicts this PR's own guarantee that hiding the enrollment entry point never prevents an existing participant from pausing or revoking.

Fix: reorder so the loading/error branches run first, and check `ui_visible` only once `status` has resolved. Worth adding a vitest case for "status fetch rejects → error + Retry still render".

### Non-blocking notes

- `_auto_upload_ui_visible` is correct: the existing-authority carve-out (`mode != "off"` / `server_enrollment_id` / `revocation_pending`) covers paused and revocation-pending states. Minor: the `_off_status` early returns (index DB missing/unreadable) can't see the enrollment row, so a corrupted DB degrades an enrolled user to flag-only visibility. Acceptable, just noting it.
- The `icacls` rewrite is sound: strip inheritance → remove non-current-user ACEs → `/grant:r` → verify protection + exclusive FullControl, with exit-code checks at each step. The old `SetOwner` call was the likely failure cause. Caveat for the record: Linux CI can't exercise this PowerShell, so we're relying on the manual Windows validation documented here.
- Surfacing PowerShell stderr in `CredentialStoreError` instead of discarding it is a clear improvement.
- Test coverage (hidden-by-default, flag-on, enrolled-stays-visible, panel/offer gating) is solid apart from the missing error-state case above.

Also a reminder for anyone following along: this merges into `feat/recurring-auto-upload-v1`, so it reaches `main` via #115 — final review happens there.

kai-rayward:
Pushed `b5920ea` to resolve the blocking item from the review above: the `ui_visible` early return sat above the loading/error branches, so an enrolled user whose status fetch failed saw nothing — no error, no Retry, no path to Pause/Turn off. The rollout gate now applies only after a status has resolved, and a regression test covers the failed-fetch → Retry → controls path. Frontend typecheck and the full vitest suite (32/32) pass locally.

PR #122, opened 2026-07-21
fix(auto-upload): egress-lock liveness + runner robustness (feature-review findings)

kai-rayward:
## Summary

Runner-robustness fixes for the recurring weekly auto-upload feature (#115), from a multi-agent review of the whole feature plus follow-up review iterations on this PR itself. The core safety invariants of #115 (hold-state gate, exact-byte egress, scope-consent binding, credential isolation, cap-of-5) were all verified holding; these commits fix the liveness/robustness defects that review surfaced.

## Commits

**1. `ce495a8` — keep raw-fingerprint re-hash out of the egress lock.** The sealed→submitting transition re-hashed up to five potentially-large raw JSONL logs *inside* `auto_upload_egress_lock`, stalling every concurrent `save_config` for the full size-unbounded re-hash (the code's own comment says only the profile-hash recompute and the transition need the lock). The locked path now passes `check_raw_fingerprints=False`; the full re-hash stays immediately before the lock.

**2. `4bd5901` — detect raw change during egress-lock wait** (review finding on commit 1). Lock acquisition can block while `save_config` holds the lock; a raw append/replace during that wait left the indexed revision unchanged and nothing after acquisition would catch it. A cheap read-free stat signature (dev/inode/size/mtime_ns — the same metadata the fingerprint ledger already compared) is captured pre-lock and re-checked immediately post-lock; mismatch aborts with `ControlChanged` before egress.

**3. `94e6357` — capture the stat baseline before the final fingerprint validation** (review finding on commit 2). Captured after it, a change landing between the validated snapshot and the baseline was baked into the baseline and shipped. Now every change is caught either by the full validation (before it) or the post-lock stat compare (after it).

**4. `68294be` — throttle stuck receipt recovery; tolerate vanished candidates** (feature-review findings #2/#3).
- *Spawn storm:* the hook due-check forced `due=True` for any `submitting` share while reconcile failures never stamped backoff — a persistently failing receipt lookup relaunched a network-calling runner on every SessionStart, forever. Reconcile failures now stamp retryable backoff; the due-check honors `next_retry_at` for `submitting`; and the sealed→submitting transition clears stale pre-egress backoff in the same transaction so a crash mid-POST still reconciles promptly (only this share's own failing submit/reconcile makes the hook wait).
- *Vanished candidate:* a top-ranked session deleted between the candidate report and the size pass produced an empty selection mislabeled as non-retryable `payload_too_large` → durable `action_required` blocking all future cycles. Missing sessions are now skipped and counted separately; an empty selection is `payload_too_large` only when something was actually size-deferred, otherwise retryable `ControlChanged`.

## Tests

Every fix carries a regression test that fails on the pre-fix code (verified by reverting), including: raw-mutation-during-lock-wait and right-after-validation aborts; due-check prompt/backoff matrix for `submitting`; recovery-failure-stamps-backoff end-to-end; stale-backoff-cleared-at-transition contract; `_ranked_size_prefix` skip/count semantics; full-cycle all-candidates-vanished → `control_changed` with backoff. Full auto-upload + daemon suites: **334 passed**.

https://claude.ai/code/session_01N6PTvGavcUjcp4mZfWx5Dt

### DIFF
--- a/clawjournal/auto_upload.py
+++ b/clawjournal/auto_upload.py
@@ -2699,12 +2699,16 @@ def _submit_pending_artifact(
     # save_config for the full, size-unbounded re-hash. Only the profile-hash
     # recompute and the submitting transition need the lock for atomicity vs
     # profile writers.
-    _validate_raw_fingerprint_ledger(conn, share)
-    # Capture a cheap, read-free stat signature of the raw inputs. Acquiring the
-    # egress lock below can block for as long as a concurrent save_config holds
-    # it, and a raw append/replace during that wait leaves the indexed revision
-    # unchanged — so the profile/revision checks inside the lock would miss it.
+    # Capture the cheap, read-free stat baseline BEFORE the final fingerprint
+    # validation below, so the baseline describes — at the latest — the very
+    # snapshot validation confirms against the sealed ledger. A raw append or
+    # replace after that snapshot (including one that lands during the possibly
+    # long wait to acquire the egress lock, when the indexed revision stays
+    # unchanged) is then visible to the post-lock stat comparison. Capturing the
+    # baseline after validation would instead bake a change made in that gap
+    # into the baseline, and check_raw_fingerprints=False would let it ship.
     pre_lock_raw_stats = _raw_stat_signatures(conn, share)
+    _validate_raw_fingerprint_ledger(conn, share)
     # The protected server check above can take long enough for a local pause,
     # hold, revision, or privacy-profile edit to win. Profile writers share
     # this short cross-process boundary lock; DB policy writers atomically bump
@@ -2712,14 +2716,14 @@ def _submit_pending_artifact(
     # recompute and submitting transition have a definite order relative to
     # every supported privacy-setting mutation.
     with config_module.auto_upload_egress_lock():
-        # A raw source that changed while we waited for the lock must abort the
-        # submission: the pre-lock fingerprint validation no longer reflects the
-        # on-disk inputs. This comparison is stat-only (no file reads), so the
-        # size-unbounded raw re-hash never runs under the lock — it stays above,
-        # before the lock, which is why check_raw_fingerprints=False here.
+        # A raw source that changed after the validated snapshot — including one
+        # that lands while we wait for the lock — must abort before egress. The
+        # comparison is stat-only (no file reads), so the size-unbounded raw
+        # re-hash never runs under the lock (which is why
+        # check_raw_fingerprints=False below).
         if _raw_stat_signatures(conn, share) != pre_lock_raw_stats:
             raise ControlChanged(
-                "A selected raw trace changed while acquiring the egress lock."
+                "A selected raw trace changed after its validated snapshot."
             )
         _validate_pending_for_submission(
             conn,
--- a/clawjournal/auto_upload.py
+++ b/clawjournal/auto_upload.py
@@ -56,7 +56,12 @@ from .auto_upload_credentials import (
 )
 from .config import load_config, save_config, source_scope_sources
 from .paths import atomic_write_text, ensure_hash_salt
-from .raw_sources import RawFingerprint, RawSourceChanged, fingerprint_raw_source
+from .raw_sources import (
+    RawFingerprint,
+    RawSourceChanged,
+    fingerprint_raw_source,
+    stat_raw_source,
+)
 from .scoring.backends import resolve_backend
 from .share_flow import build_zip, package, verify_coverage
 from .workbench.index import (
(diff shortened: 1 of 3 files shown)


kai-rayward:
Follow-up commit `4bd5901` — closes a TOCTOU window that the first commit reopened.

**The gap:** moving the raw-fingerprint re-hash out of the egress lock meant the only raw validation ran *before* acquiring the lock. But `auto_upload_egress_lock` acquisition blocks for as long as a concurrent `save_config` holds it, and a selected raw log appended-to or replaced **during that wait** leaves the indexed revision unchanged — so the in-lock profile/revision checks miss it, and nothing after lock acquisition would catch the changed input.

**The fix:** capture a cheap, read-free stat signature (`dev/inode/size/mtime_ns` — the same metadata the fingerprint ledger already compared) before the lock, and re-check it immediately **after** acquiring the lock. A mismatch raises `ControlChanged` and aborts before egress. This detects a lock-wait raw change without paying the size-unbounded re-hash under the lock, so the liveness win of the first commit stands.

- `raw_sources.stat_raw_source()` — stat-only signature for a file or subagent directory; a vanished input counts as a change.
- `auto_upload._raw_stat_signatures()` — per-session signatures for a sealed share.
- `_submit_pending_artifact` — capture pre-lock, re-check post-lock inside the lock.

New tests: a raw append during a mocked lock acquisition aborts `_submit_pending_artifact` with `ControlChanged` and leaves the share `sealed`; the stat signature changes on both an append and a same-size replace. Full auto-upload + raw_sources suites green.

kai-rayward:
Follow-up commit `94e6357` — fixes an ordering bug in the lock-wait guard from `4bd5901`.

**The gap:** the stat baseline was captured *after* the final `_validate_raw_fingerprint_ledger`. A raw append/replace landing between the validated snapshot and the baseline capture would be baked into the baseline itself — so the post-lock comparison matched, and with `check_raw_fingerprints=False` the stale sealed artifact could still ship (the indexed revision stays unchanged, so the in-lock profile/revision checks miss it too).

**The fix:** capture the baseline immediately *before* the final validation, so the validated snapshot is the latest state the baseline can describe. Any change after that snapshot is caught either by the validation (if before it) or by the post-lock stat comparison (if after it) — the full pre-lock → lock-wait window is now covered end to end.

Regression test `test_submit_aborts_when_raw_changes_right_after_validation` mutates the raw log the instant the final pre-lock validation returns; it aborts with `ControlChanged` under the fixed order and reaches `submit_artifact` under the old order (confirmed by temporarily reverting the two lines). Full auto-upload + raw_sources suites green (155 passed).

kai-rayward:
## Adversarial review round on this PR

Ran two independent adversarial review passes over the full PR diff (concurrency/state-machine lens + regression/contract lens). Both converged on one real **MAJOR** regression introduced by the storm-guard commit `68294be`, fixed in `0b23046`:

**The recovery-branch failure stamps clobbered durable status overlays.** `_record_cycle_result(retryable=True)` rewrites `health='retrying'` + `last_result_code`, but receipt recovery deliberately runs *before every lifecycle gate* — on paused/off/`action_required` enrollments. Worst case (fully organic): ambiguous POST → user edits a privacy setting (`paused` + `action_required`, fresh generation) → one transient lookup failure erases `action_required` → the later successful receipt commit takes the non-preserve branch → **automatic egress silently re-opens past a safety stop the user never reviewed**. Fix: a dedicated `_record_recovery_backoff` stamps only `consecutive_failures` + `next_retry_at` (all the hook throttle reads) and never touches health/result codes. Also paced the success-shaped `"disabled"` probe path (receipt_not_found while Off left the share `submitting` with nothing stamped — leftover hooks could still storm), reset the failure exponent at the sealed→submitting clear, and fixed a stale return annotation.

**Verified sound by the review** (beyond the fixed finding): the full submit-path TOCTOU interval enumeration (every window from sealed-ledger validation to POST is covered; the sealed ZIP's sha256 is re-verified at the last moment inside `submit_artifact`); `stat_raw_source` blind-spot analysis (nanosecond mtime on APFS/ext4; dir-member semantics consistent with the hash path); transition-clear atomicity/generation-guards; due-check starvation (none — a `submitting` share always becomes due again); `_ranked_size_prefix` skip semantics vs downstream snapshot/fingerprint binding; no stale 2-tuple unpackers; frontend `exclusion_counts` unaffected.

**Known-minor left open** (documented, not fixed here): terminal-but-throttled recovery failures (wiped credential store, foreign-enrollment share) retry forever at the 24h cap with no escalation to a user-facing action — the design currently can't express "terminal + throttled" for a `submitting` share without resurrecting the storm; post-size-break vanished candidates are attributed to `deferred_by_size` (reporting accuracy only); Windows `whole_run_lock` file grows one byte per acquisition (cosmetic); hook `run` path cold-imports the heavy stack each SessionStart (bounded by the 5s timeout).

Full auto-upload + daemon suites: **337 passed**.

PR #123, opened 2026-07-21
feat(auto-upload): speak hosted recurring protocol v2

kai-rayward:
## Summary

Implements **hosted recurring-upload protocol v2**, unblocking auto-upload enrollment: the hosted service moved to v2 with its logical-session-revisions rollout, so the v1 client failed capability validation (`capability_incompatible`) and could never enroll.

## What v2 requires

1. **Explicit scope entries** — enrollment sends the exact authorized `[{source, project}…]` pairs instead of a client-computed `scope_hash`. These entries are the **full cross product of the confirmed sources × projects** — precisely the set the candidate filter enforces (`source ∈ sources ∧ project ∈ projects`) and that both consent surfaces display, so the server-certified scope always covers exactly what the runner can select (never a narrower "only-observed-so-far" subset).
2. **Ownership certification** — a versioned certification the user must see and accept as a distinct affirmative act.
3. **Server-owned scope hash** — the server computes the HMAC; the client pins the read-back value and fails closed on drift.

## Changes

- `auto_upload_client.py`: v2 constants and payloads; ownership-certification guard; mutating transport, truncation, 5xx, and malformed-body failures are classified as ambiguous.
- `auto_upload.py`: challenge/profile binding, server scope-hash read-back, and fail-closed recovery for ambiguous PATCH/create/credential-rotation outcomes. PATCHes compensate with known recovery authority; lost email-token POSTs require fresh verification and recover the same stable enrollment intent. Resume/Disable refuse stale recovery credentials until that reissue completes.
- `workbench/index.py`: gated migration v9 for ownership certification and server scope hash.
- CLI and Workbench: show every exact source/project pair before consent; malformed pair lists are rejected rather than partially displayed; ownership remains a separate affirmative act.
- Manual-share success now caches recurring-offer availability against the shared protocol-v2 constant instead of stale v1.
- Hosted capability/terms checks run before the strict source scan, with a CLI progress notice.
- Docs updated to the v2 contract.

## Verification

- Full Python suite: **2,771 passed** (adds an end-to-end daemon test that a successful v2 manual share caches `auto_upload_capability_available=True`, so the Workbench enable offer appears).
- Frontend: **34 passed**; changed-file ESLint clean; production build succeeded.
- Added adversarial coverage for ambiguous PATCH compensation, first-create recovery, credential-rotation reissue using the same client ID, unsafe Resume/Disable blocking, malformed consent scope, v2 offer caching, and mutating-response ambiguity.
- Live PR head was guarded before push; fix commit: `35f7509`.

https://claude.ai/code/session_01N6PTvGavcUjcp4mZfWx5Dt


kai-rayward:
## Adversarial review round (two lenses) — findings fixed in `fd44c18`

**Consent/authorization lens + lifecycle/compat lens**, both over the full diff; all findings verified against the code before fixing.

**MAJOR — certified scope ≠ enforced scope.** Enrollment sent only the (source, project) pairs *observed* at enable time, but the candidate filter admits the full cross product of enrolled sources × projects — a later session in a new combination of already-enrolled source and project would egress outside the server-certified pair list. Entries are now the **full cross product**, which is also exactly what the consent dialog displays (a sources list and a projects list). Certification, enforcement, and display now agree.

**MAJOR — oversized scopes failed late and server-worded.** The hosted contract caps scope entries at 200. A `scope_too_large` blocker now fails fast with an actionable message (exclude projects / narrow sources) before consent, network, or the multi-minute strict scan.

**MAJOR — fresh-token loss window.** The post-create read-back GET sat between a credential (re)issue and any persistence: a transient failure dropped the freshly issued tokens, and if the reissue had invalidated the prior recovery token server-side, `disable()` would retry revocation with a dead token forever (unrecoverable without DB surgery). The read-back now runs **after Phase-1**, so the newest recovery token is always durably persisted before the extra network step — a failed or crashed read leaves revocable recovery-only authority. Regression test pins it: read-back *and* compensating revoke both failing still retains the fresh recovery tombstone + `revocation_pending`.

**Minors fixed:** distinct `reauthorization_required` code when the scope-hash pin is missing (pre-v2 rows) instead of implying hosted drift; the CLI scan notice prints before *every* `enable()` rerun; `verifyCode()` names the unchecked certification box instead of silently no-oping after the code is consumed; a v1-shaped 409 from an older running daemon renders a restart-the-service explanation instead of an unwinnable retry loop; dead v1 `scope_hash()` deleted.

**Reviewed sound** (one line each): consent cannot reach the network without all four exact-version acceptances on any surface; profile hash binds the certification and entries; v1-enrollment upgrade paths fail closed with no retry storm (pause/resume/disable/receipt-recovery all safe on v1-era rows); migration v9 idempotent across fresh/v8/pre-v7; no stale callers of the old client signatures; two-checkbox gating consistent across all frontend paths.

**Deliberately not changed:** the client keeps exact `recurring_upload_api_version == 2` (no dual-version window) — the flag-day trade-off is a server-side rollout concern; the capability doc's `supported_recurring_client_versions` list is where overlap belongs.

Suites after fixes: **2,750 Python + 33 frontend passed**, dist rebuilt.

kai-rayward:
Follow-up commit `188c583` — closes a consent-integrity TOCTOU in the enrollment flow (external review finding).

**The gap:** the enrollment POST/PATCH carries only `ownership_certification: true` — the server records whatever certification version is *current at request time*. A rotation in the [fetch_authorization → POST] window meant the client committed Enabled while pinning the stale challenge version: authority active under terms the user never reviewed, plus a pin the runner gate could never match (permanent mismatch until reauthorization). The authorization/retention versions don't have this hole — the request sends them and the server rejects stale values; the certification boolean is the only consent field the server can't validate against what the user actually saw.

**The fix:** the post-Phase-1 read-back now compares the recorded `ownership_certification_version` against the exact accepted version and aborts fail-closed on mismatch, before any commit; the create path's compensating revoke removes the just-created enrollment. Regression test pins abort + revoke + nothing-committed, and fails with the guard removed (revert-verified). Full suite: **2,751 passed**.

kai-rayward:
Follow-up commit `7bb7e98` — external review finding on `188c583`: the certification read-back guard was incomplete on the **update path**.

**The gap:** when the certification rotates during an existing-enrollment PATCH, the guard fires only after the server has already applied the new authorization revision — and the generic updating-failure branch merely paused local state without ever calling `revoke_enrollment`. The hosted enrollment stayed **active under a certification version the user never accepted**: no egress (the local gate fails closed), but the server-side consent record misrepresented the user, and the client cannot un-PATCH.

**The fix:** `certification_not_accepted` (now its own code) routes the update path into the same revoke machinery the create path uses — revoke with recovery authority + Off cleanup, or a `revocation_pending` recovery tombstone if the revoke itself fails so Disable can finish. All other updating failures still preserve the legitimate prior enrollment and pause. Parametrized regression tests cover both revoke outcomes and are revert-verified (fail with the routing removed). Full suite: **2,753 passed**.

kai-rayward:
Follow-up commit `142ca79` — external review finding on `7bb7e98`: the revocation predicate was still too narrow.

**The gap:** after a definite PATCH, the certification boolean gives the client no way to know which certification version the server recorded — only the read-back can establish it. The predicate routed only the *explicitly confirmed* `certification_not_accepted` mismatch into revocation; a failed read-back GET or a malformed response (no scope hash) took the generic updating branch and merely **paused** a live hosted enrollment whose consent record cannot be shown to match what the user accepted.

**The fix:** `enable()` tracks `certification_verified` — set only when the read-back confirms the exact accepted version — and routes **every** updating failure between the definite PATCH and that verification into the compensating revoke: explicit mismatch, transport failure, malformed read-back, or a control race in the window. Failures after verification (or before the PATCH was definite) still preserve the legitimate enrollment and pause. The ControlChanged pause-merge also now requires the verification flag explicitly rather than by statement adjacency.

Parametrized regression tests (GET failure / missing scope hash) assert revoke + Off cleanup + the original error surfaced, and are revert-verified. Full suite: **2,755 passed**.

kai-rayward:
Follow-up commit `d195bc0` — external review P1 on the certification guard: a malformed **2xx** PATCH body slipped past revocation.

**The gap:** response-body validation (missing `enrolled_at`/`authorization_revision`, an id change, a boundary moved later) raised *before* `server_reauthorization_succeeded` was set and before the read-back — so the revoke predicate was false and the generic update path merely **paused**, leaving a definitely-applied PATCH live under a certification the client never verified. The revoke machinery also keyed off `response_local["enrollment_id"]`, which a malformed body may not even contain.

**The fix, exactly as flagged:**
- Mark the server mutation definite the instant `create`/`update_enrollment` returns 2xx — *before* body validation.
- Derive the updating-path revoke identity from **known** state — the existing `server_enrollment_id` we PATCHed and the pinned recovery credential (or the reissued token when it validated) — never the response body.

Every updating failure between a definite PATCH and a verified certification now revokes with that identity; a failed revoke leaves the `revocation_pending` recovery tombstone for Disable.

**Behavior change (documented):** the boundary-moved-later reauthorization case now revokes instead of pausing — it's a definite PATCH whose certification couldn't be verified, so paused-but-live violated the read-back consent contract. Its test is updated to the revoked end-state and still asserts the rejection.

Parametrized malformed-body tests (missing revision / enrolled_at / id / empty) all revoke with the known id + pinned recovery and end Off with credentials cleared; revert-verified against the early definiteness marker. Full suite: **2,759 passed**.

JustMeStudio:
Thanks — I reviewed the latest commit (`9d8ff09`). The core v2 enrollment, ownership certification, v9 migration, server-owned scope pinning, and compensation paths look solid.

I found one blocking v1 remnant:

`clawjournal/workbench/daemon.py` still caches `auto_upload_capability_available` only when `recurring_upload_api_version == 1`.

The live service now returns version 2, so after a successful manual share the client stores the capability as unavailable. Consequently, `offer_available` remains false and the Workbench enable offer stays hidden, although direct CLI enrollment can work.

Suggested fix: compare against `RECURRING_UPLOAD_API_VERSION` or reuse the v2 capability validator, and add a daemon test proving that a successful manual share against v2 makes the offer available.

One scope question: the PR summary says enrollment sends observed `(source, project)` pairs, but `_current_scope()` currently sends the full source × project Cartesian product. If that is intentional, the PR description/consent wording should say so; otherwise candidate filtering and persisted scope should use exact pairs.

Validation performed:
- Live `data.rayward.ai` v2 capability validation passed
- 331 focused Python tests passed
- 33 frontend tests passed
- Production frontend build passed

I recommend fixing the capability cache before merging.

kai-rayward:
Thanks for the careful pass — appreciate the live-service validation especially. Two responses:

### 1. Capability cache `== 1` — this is already fixed on the branch (I think you were reading pre-v2 code)

`daemon.py` no longer hardcodes `recurring_upload_api_version == 1`. That check was removed in this PR and replaced with `_recurring_offer_available()`, which compares against the shared `RECURRING_UPLOAD_API_VERSION` constant (now `2`). Evidence:

- **Static**: `_recurring_offer_available()` → `capabilities.get("recurring_upload_api_version") == RECURRING_UPLOAD_API_VERSION` (imported from `auto_upload_client`, value `2`).
- **Historical**: the hardcoded `== 1` lived at `daemon.py:4567` on the *old* main (`d18ddc1`); this PR's feat commit deleted it. If you diffed only the latest commit, the daemon change isn't in that range.
- **Unit test**: `test_recurring_offer_uses_current_protocol_version` asserts v2 → `True`, v1 → `False`, closed → `False`.
- **Runtime**: against the live `data.rayward.ai` v2 capability document, `_recurring_offer_available()` returns `True` — so a successful manual share caches `auto_upload_capability_available=True` and the offer appears.

Per your suggestion I also added the **end-to-end** guard you asked for — `test_successful_v2_manual_share_caches_offer_available` drives the share-success handler and asserts it writes the offer flag `True` for a v2 capability doc. Revert-verified: reintroducing `== 1` fails both this and the unit test.

### 2. Scope: the cross product is intentional, and certified == enforced == displayed

`_current_scope()` deliberately certifies the **full sources × projects cross product**, because that is exactly what the candidate filter enforces (`source ∈ sources ∧ project ∈ projects`). Certifying only the pairs observed at enrollment would let a later trace in a new combination of already-authorized source and project egress outside the certified set — that was an earlier review finding, fixed here. So certified scope, enforced scope, and what's shown at consent are all the same set. Both surfaces render every exact pair (Workbench: "Exact recurring scope · N source/project pairs"; CLI: "Exact authorized source/project pairs:"). I've also clarified the PR description to state the cross-product explicitly.

Net: I don't believe #1 is a blocker for merge — happy to walk through any of the above if something still looks off.

PR #124, opened 2026-07-21
fix(auto-upload): migrate legacy session hooks

JustMeStudio:
## Summary

Fix automatic-upload Hook migration for users upgrading from the earlier recurring-upload implementation.

Legacy installations used a `clawjournal.cli auto-upload hook` handler marked with `clawjournalProfile: clawjournal-auto-upload-v1`. The current installer only recognized `clawjournal.agent_hooks run`, leaving both handlers installed after an upgrade.

## Changes

- Recognize the exact legacy ClawJournal Hook profile marker.
- Replace legacy handlers with the current SessionStart handler during installation.
- Collapse legacy and current duplicates into one handler.
- Remove both legacy and current handlers when automatic upload is disabled.
- Preserve unrelated third-party hooks.
- Keep repeated installation and removal idempotent.

## Validation

- Added legacy upgrade and uninstall regression coverage.
- `106 passed` across Hook, automatic-upload, and CLI tests.
- `git diff --check` passed.

### DIFF
--- a/clawjournal/agent_hooks.py
+++ b/clawjournal/agent_hooks.py
@@ -161,13 +161,23 @@ def _handler_for(client: AgentName) -> dict[str, Any]:
     return handler
 
 
-def _handler_is_ours(handler: Any) -> bool:
+def _handler_is_current(handler: Any) -> bool:
     if not isinstance(handler, dict) or handler.get("type") != "command":
         return False
-    if handler.get("clawjournalProfile") == LEGACY_HOOK_PROFILE:
-        return True
     command = str(handler.get("command", ""))
     return f"-m {HOOK_MODULE}" in command and " run " in f" {command} "
+
+
+def _handler_is_legacy(handler: Any) -> bool:
+    return (
+        isinstance(handler, dict)
+        and handler.get("type") == "command"
+        and handler.get("clawjournalProfile") == LEGACY_HOOK_PROFILE
+    )
+
+
+def _handler_is_ours(handler: Any) -> bool:
+    return _handler_is_current(handler) or _handler_is_legacy(handler)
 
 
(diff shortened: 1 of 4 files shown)


kai-rayward:
Reviewed the migration and pushed one fix on top (66fac90).

**The migration itself checks out.** Verified against the legacy installer in history (`auto_upload_scheduler.py` from the original recurring-uploads branch): the marker value and handler shape match exactly, install replaces/collapses legacy handlers idempotently, uninstall removes them, and foreign hooks — including ones carrying a different `clawjournalProfile` — are preserved.

**One regression, fixed in 66fac90:** broadening `_handler_is_ours` also changed `hook_diagnostics`, where `configured` requires *exactly one* matching handler. For the upgrade population this PR targets — a legacy handler still sitting next to a healthy current hook, with nothing re-running install until the user manually re-runs `enable` — the count became 2, so `configured` flipped to `False`. Downstream, `_selected_hook_missing()` then blocks every scheduled cycle with `hook_missing` and status shows `action_required`. Repro: the identical hooks file gives `configured=True` on main but `False` on this branch as of fda5f3b. Net effect would have been silently stopping weekly auto-uploads for exactly the users being migrated.

The fix splits the predicate into `_handler_is_current` / `_handler_is_legacy` (install/uninstall still sweep both), computes `configured` from current handlers only, and surfaces the leftover as `legacy_hook_installed` in diagnostics plus a one-line hint in `auto-upload status` (`enable` migrates, `disable` removes — disable sweeps hooks even when not enrolled). The workbench frontend whitelists status fields, so no frontend change is needed.

Added three regression tests (legacy+current coexistence stays configured and collapses on install; legacy-only reads installed-but-unconfigured; status hint prints and is sanitized). Full suite: 2749 passed; CI is green on 66fac90.

PR #125, opened 2026-07-21
feat(redaction): Betterleaks scanner foundation (PR 1/4)

kai-rayward:
## Summary

First of four PRs replacing the all-or-nothing TruffleHog share gate with a tiered per-finding policy (**Betterleaks primary detection + TruffleHog verified-only secondary**). Context: on a real 33-session auto-upload pool, the current gate sent 17 sessions to `pending_review` because *any* finding (verified or not) blocks the whole bundle.

This PR is **pure addition — no behavior change**: nothing is wired into the share gate yet. It lands the scanner wrapper, managed install, bundled config, CLI verb, and tests.

- `clawjournal/redaction/betterleaks.py` — subprocess wrapper mirroring `trufflehog.py`'s surface: `scan_file`/`scan_text`, raw-match internals (`_scan_file_for_raw_matches` for the future gate redact-loop, `_scan_text_for_raw_matches` for the findings engine), `scan_session_for_betterleaks_findings` + `betterleaks_secret_map_from_blob` (the same two-function contract TruffleHog implements), managed-binary resolution, memoized `engine_fingerprint`, `managed_off_pin`.
- `clawjournal/redaction/betterleaks.toml` — bundled share-gate config: `[extend] useDefault = true` plus an allowlist for our own `[REDACTED_*]` placeholders (so the future redact-and-rescan loop converges). Shipped via `package-data`.
- `betterleaks_install.py` + `scripts/check_betterleaks_pin.py` + `.github/workflows/betterleaks-pin.yml` — managed install pinned to **v1.6.1** with vendored release checksums; mirrors the TruffleHog pin machinery. Upstream naming quirks handled: `x64` (not `amd64`) arch keys, `.zip` archives on Windows.
- `clawjournal betterleaks install|status` CLI verb (clone of the trufflehog one).
- `tests/conftest.py` — autouse `CLAWJOURNAL_SKIP_BETTERLEAKS=1` beside the TruffleHog bypass.

## Privacy posture

- Live validation (`--validation`) is **never** passed — candidate secrets must not leave the machine from the detection layer; verification remains TruffleHog's role (unchanged, and narrowed to verified-only in PR 2).
- `--redact` is not used because it also scrubs the report; instead the raw `Secret` is parsed transiently (0600 mkstemp report file, read + unlinked in `finally`) exactly like the TruffleHog wrapper parses `Raw` from stdout. Raw values never leave the module via public helpers; masked + salted-hash only.
- Subprocess env is scrubbed via the same `_scrubbed_subprocess_env` allowlist.

## Verified against the real binary (1.6.1)

- Report is a JSON **array** (`null` on clean scans); fields `RuleID`/`StartLine`/`Secret`/`Match`/`Entropy`/`Fingerprint` — a real report is vendored as `tests/redaction/fixtures/betterleaks_report.json` (Slack-token value defanged post-capture: GitHub push protection rightly rejects a realistic one).
- `--exit-code 183` honored → both wrappers share the `{0, 183}` = "scan ran" contract.
- `StartLine` is 1-based and aligns with `sessions.jsonl` line = session index (the mapping the gate uses).
- A `\n`-escaped private key inside a JSONL string **is** caught, and `Secret` arrives in file-level escaped form — this is why the PR 2 gate loop will string-replace on serialized lines rather than object-walk.
- `stdin` mode supports `--report-path` and line numbers; version banners parsed for both `--version` and `version` forms.

## Tests

87 new tests (wrapper subprocess contract driven by the vendored real report; install matrix incl. zip flavor, checksum fail-closed, traversal rejection; pin-check script; CLI dispatch). Full suite: **2858 passed**.

## Next PRs

2. `scan_policy.py` + `share_gate.py` redact-and-rescan loop + both gate call-site rewrites.
3. Findings-engine swap (config default, fingerprints, apply wiring, preview gates).
4. Auto-upload per-tier routing + docs (CLAUDE.md invariant rewrite, PRIVACY.md, README).

### DIFF
--- a/.github/workflows/betterleaks-pin.yml
+++ b/.github/workflows/betterleaks-pin.yml
@@ -1,0 +1,66 @@
+name: Betterleaks Pin Integrity
+
+# Verifies that the sha256 checksums vendored in
+# clawjournal/redaction/betterleaks_install.py still match the upstream
+# release they pin. Unit tests are hermetic and cannot see this — only a
+# real fetch of upstream's checksums.txt catches a bad pin bump or a
+# release asset that was silently re-uploaded.
+#
+# IMPORTANT: do NOT add this workflow's check to branch-protection
+# required checks. It is path-filtered at the trigger level, so on PRs
+# that do not touch the pinned files it produces no check run at all — a
+# required status would then sit "Expected" forever and block the PR.
+
+on:
+  pull_request:
+    paths:
+      - 'clawjournal/redaction/betterleaks_install.py'
+      - 'scripts/check_betterleaks_pin.py'
+      - '.github/workflows/betterleaks-pin.yml'
+  schedule:
+    # Mondays 06:23 UTC — off the top of the hour, which GitHub delays
+    # under load, and offset from trufflehog-pin.yml's 06:17. Note:
+    # scheduled runs fire only from the default branch (so this
+    # activates after merge), and GitHub auto-disables crons on public
+    # repos after 60 days with no repository activity. Cron failure
+    # emails go to whoever last edited this file.
+    - cron: '23 6 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: betterleaks-pin-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  pin-integrity:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Fetch upstream checksums.txt (authenticated, retried)
+        env:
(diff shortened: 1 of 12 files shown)


kai-rayward:
**Adversarial review: merge-ready.** All four security invariants verified by an independent reviewer (no `--validation` ever passed; raw plaintext transient with 0600 mkstemp + unlink-in-finally; scrubbed subprocess env; traversal-safe pinned extraction with checksum-before-execute). Two follow-ups landed:

- Confirmed against the real 1.6.1 binary that a clean scan writes `null` (exit 0), which `_read_report_payload` maps to an empty findings list — the fail-closed-on-clean scenario the reviewer flagged cannot happen with the pinned version.
- The `_scan_text_for_raw_matches` docstring overstated the SIGKILL guarantee (the JSON *report* tempfile can still be orphaned in the write→unlink window; only the *session* payload is never written to disk). Docstring now says exactly that.

PR #126, opened 2026-07-21
feat(redaction): tiered secret-scan share gate (PR 2/4)

kai-rayward:
## Summary

Stacked on #125. This is the behavior change: the all-or-nothing TruffleHog gate becomes a **per-finding tier policy** — Betterleaks primary detection (local-only, no `--validation` ever), TruffleHog narrowed to `--results=verified` as the live-credential check, and the gate **redacts instead of rejecting** wherever that is safe. No third-party scanner decides whole-session fate anymore.

## Tier policy (`scan_policy.classify`, precedence order)

| Signal | Tier |
|---|---|
| TruffleHog-verified live credential | **block** (allowlisted-but-verified → **review** — a confirmed-live credential never ships silently) |
| Value the user ignored/allowlisted (findings-table decision) | **warn** — fixes the long-standing trap where allowlisting made the redactor skip a value and the gate then re-found and blocked it |
| Private-key / pkcs12 structure | **review** |
| Soft rules (`generic-api-key`), low-entropy hits | **warn** |
| Recognizable-but-unverified token (the common case) | **redact** in place; share proceeds |

Scanner failure/missing binary still fails closed (`scanner-error` / `scanner-not-installed`); any bypass env var marks the whole gate bypassed — recorded in the manifest, and the upload path refuses to ship it (`secret-scan-bypassed`).

## Redact-and-rescan loop (`workbench/share_gate.py`)

- Each pass: one Betterleaks + one TruffleHog subprocess (`scan_file_with_raws` — same scan also yields the raw matches, no double scanning), classify against the mapped session's findings-table decisions, rewrite, rescan; bounded at 3 passes, non-convergence escalates the stragglers to review.
- Rewrites are **string-level on the serialized JSONL line** (scanners report on-disk escaped bytes — a `\n`-escaped private key arrives escaped; verified against the real binary in PR 1) with `json.loads` re-validation; a span that would corrupt a line is rolled back and escalated (`unredactable_span`).
- Block/review raws are scrubbed from the on-disk debug artifact too; blocking is decided by the accumulated findings, not by what remains in the file. All rewriting completes before callers hash/seal — the sealed-bytes ledger invariant is untouched.

## Plumbing

- Both gate call sites rewired; `finalize_share_export_for_upload` gains a `conn` param (decisions lookup).
- New zip artifacts `secret-scan.json` / `secret-scan.post-pii.json` alongside the legacy `trufflehog*` names, which now show all-clean on every shippable bundle (verified findings are block-tier) — hosted-side consumers of the legacy artifact see the same shape as today.
- `_manifest_is_finalized_for_upload` is tier-aware (warn-tier manifests are finalized; block/review/non-converged are not); pre-tier cached exports rebuild once.
- `blocked_sessions` entries keep the legacy `detector` key and gain `engine`/`rule`/`tier`; share_cli remove-and-retry and the auto-upload pending_review routing work unchanged (full per-tier runner handling lands in PR 4).
- Frontend: `BlockedShareFinding` gains optional `engine`/`rule`/`tier`; Package step renders a tier chip. `npm run build` verified.

## Verification

- **Real-binary end-to-end** (betterleaks 1.6.1 + trufflehog 3.95.3, bypasses unset): fake Slack token → auto-redacted, session ships; private key → review, only its session blocks; 2-pass convergence; every JSONL line still valid; no raws left on disk. This is the "29 clean, 3 redacted, 1 review" shape the 33-session incident should have produced.
- 68 new tests (`test_scan_policy.py` tier table + report semantics, `test_share_gate.py` rewrite mechanics incl. escaped-form and JSON-rollback, `TestSecretScanGate` end-to-end tier behavior — including the inverted assertion: unverified finding now advances share status).
- Full suite: **2900 passed**.

### DIFF
--- a/.github/workflows/betterleaks-pin.yml
+++ b/.github/workflows/betterleaks-pin.yml
@@ -1,0 +1,66 @@
+name: Betterleaks Pin Integrity
+
+# Verifies that the sha256 checksums vendored in
+# clawjournal/redaction/betterleaks_install.py still match the upstream
+# release they pin. Unit tests are hermetic and cannot see this — only a
+# real fetch of upstream's checksums.txt catches a bad pin bump or a
+# release asset that was silently re-uploaded.
+#
+# IMPORTANT: do NOT add this workflow's check to branch-protection
+# required checks. It is path-filtered at the trigger level, so on PRs
+# that do not touch the pinned files it produces no check run at all — a
+# required status would then sit "Expected" forever and block the PR.
+
+on:
+  pull_request:
+    paths:
+      - 'clawjournal/redaction/betterleaks_install.py'
+      - 'scripts/check_betterleaks_pin.py'
+      - '.github/workflows/betterleaks-pin.yml'
+  schedule:
+    # Mondays 06:23 UTC — off the top of the hour, which GitHub delays
+    # under load, and offset from trufflehog-pin.yml's 06:17. Note:
+    # scheduled runs fire only from the default branch (so this
+    # activates after merge), and GitHub auto-disables crons on public
+    # repos after 60 days with no repository activity. Cron failure
+    # emails go to whoever last edited this file.
+    - cron: '23 6 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: betterleaks-pin-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  pin-integrity:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Fetch upstream checksums.txt (authenticated, retried)
+        env:
+          # Authenticated so we get this repo's GITHUB_TOKEN rate limit
+          # rather than the shared runner-IP unauthenticated limit.
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          PINNED=$(python -c "from clawjournal.redaction.betterleaks_install import PINNED_VERSION; print(PINNED_VERSION)")
+          echo "Pinned Betterleaks version: v$PINNED"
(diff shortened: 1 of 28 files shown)


kai-rayward:
**Adversarial review (4 parallel reviewers, one per PR in the stack) found two real defects here — both fixed in `fix(share-gate): raw-less finding backstop + per-line redaction scoping`:**

1. **Raw-less finding bypass (major).** `scan_file_with_raws` only surfaces raw matches with a non-empty secret value, but both parsers still count the finding in their report. The gate tiered only the raw matches, so a TruffleHog-**verified** finding whose `Raw` was empty/absent never blocked — worst case, a live credential ships. Now any report finding with `raw_sha256=None` becomes a sticky policy finding (`finding_without_raw`): verified → block, everything else → review. Regression test: `test_finding_without_raw_blocks_instead_of_vanishing`.

2. **Cross-session warn-tier stripping (major).** `_rewrite_sessions_file` applied one flat raw→placeholder map to every line, so a redact-tier value in session B also got stripped from session A where the user's ignore decision made it warn-tier — silently discarding data the user chose to keep, and making the manifest contradict the bytes. Redaction is now scoped to each finding's line (line-less findings still apply globally, erring toward redaction), and JSON-rollback failures are tracked per `(line, raw)`. Regression tests: `test_redact_tier_does_not_strip_ignored_value_in_other_session`, `test_redaction_is_scoped_to_the_findings_line`.

Minor items also addressed: docstrings no longer overstate the on-disk scrub of block/review raws (an unredactable span stays in the *blocked* export dir), and `_DecisionCache` documents the escaped-form hash contract.

**Hosted-compat question from the PR body is now resolved with evidence:** `clawjournal-share`'s `validation.py` requires `trufflehog.post-pii.json` (JSON object) in the zip and `redaction_summary.trufflehog_post_pii` with `findings == 0`, `bypassed === false`, `binary_missing === false`, falsy `scan_error`. The finalize path writes exactly that shape from the verified-only sub-report on every shippable bundle — no server change needed.

PR #127, opened 2026-07-21
feat(findings): betterleaks becomes a default findings engine (PR 3/4)

kai-rayward:
## Summary

Stacked on #126. Wires the share gate's primary scanner into the scan-time findings pipeline and the preview gates, so the local review UI surfaces the same findings the gate will act on at share time.

- **Engine default swap**: `get_enabled_engines` default is now `(betterleaks, regex_pii, regex_secrets)`; the trufflehog engine remains an explicit opt-in. A config migration appends `betterleaks` to explicit `enabled_findings_engines` lists while preserving user opt-ins; absent keys ride the default.
- **Config formalization**: `enabled_findings_engines` joins `ClawJournalConfig`, `_AUTO_UPLOAD_PROFILE_CONFIG_KEYS` (changing the engine set pauses auto-upload enrollment — it shapes what leaves the machine), `get_effective_share_settings`, and `_redaction_settings_fingerprint` (toggling an engine invalidates finalized bundles).
- **Revision fold**: `compute_findings_revision` folds the betterleaks `engine_fingerprint` like the trufflehog one.
- **Pipeline + apply**: fail-soft betterleaks branches in `run_findings_pipeline` and the deterministic redaction-log builder; `apply_findings_to_blob` hoists `betterleaks_secret_map_from_blob` beside the trufflehog map (one subprocess per apply).
- **Preview gates**: events bundle export runs both scanners through the new `scan_policy.preview_gate` (tier-aware — warn passes; `block_reason` becomes `secret-scan-findings`); the rendered-skill gate gets a tier-aware betterleaks section; skill preflight requires the betterleaks binary.

## ⚠️ One-time cost

The default-engine change + new fingerprint flips `findings_revision` for **every session** → one full findings rescan after upgrade, bounded by the existing settle/backfill machinery.

## Tests

New: migration table (`TestMigrateFindingsEngines`), betterleaks pipeline rows + fail-soft crash test, revision-fingerprint participation, fingerprint-matrix extension. Updated: engine-default expectations, events gate block-reason. Full suite: **2909 passed**.

### DIFF
--- a/.github/workflows/betterleaks-pin.yml
+++ b/.github/workflows/betterleaks-pin.yml
@@ -1,0 +1,66 @@
+name: Betterleaks Pin Integrity
+
+# Verifies that the sha256 checksums vendored in
+# clawjournal/redaction/betterleaks_install.py still match the upstream
+# release they pin. Unit tests are hermetic and cannot see this — only a
+# real fetch of upstream's checksums.txt catches a bad pin bump or a
+# release asset that was silently re-uploaded.
+#
+# IMPORTANT: do NOT add this workflow's check to branch-protection
+# required checks. It is path-filtered at the trigger level, so on PRs
+# that do not touch the pinned files it produces no check run at all — a
+# required status would then sit "Expected" forever and block the PR.
+
+on:
+  pull_request:
+    paths:
+      - 'clawjournal/redaction/betterleaks_install.py'
+      - 'scripts/check_betterleaks_pin.py'
+      - '.github/workflows/betterleaks-pin.yml'
+  schedule:
+    # Mondays 06:23 UTC — off the top of the hour, which GitHub delays
+    # under load, and offset from trufflehog-pin.yml's 06:17. Note:
+    # scheduled runs fire only from the default branch (so this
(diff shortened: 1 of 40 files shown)


kai-rayward:
**Adversarial review: merge-ready, no blocking code defects.** Verified correct by an independent reviewer: idempotent migration, order-independent fingerprints (no spurious finalized-bundle invalidation), fail-soft pipeline branch, one-subprocess-per-blob apply path, bounded backfill. Landed since: unit tests for `preview_gate` composition semantics (warn passes / redact+review block / any-TruffleHog-finding blocks / missing-binary + scan-error block / bypassed engine contributes nothing), a betterleaks-missing preflight test, and events-bundle blocked manifests now carry a `block_message` with the install command when the reason is `scanner-not-installed`.

Two deliberate behaviors worth sign-off (unchanged): (1) the preview gate blocks event/skill exports on redact-tier findings with no in-path remediation — fail-closed by design for derived digests; (2) enrollment-pause reads the *raw* config engine list while bundle invalidation reads the *resolved* set, so default-riding users get bundles rebuilt (stronger redaction) without an enrollment pause on upgrade.

PR #128, opened 2026-07-21
feat(auto-upload): per-tier finding routing + tiered-gate docs (PR 4/4)

kai-rayward:
## Summary

Stacked on #127. Completes the switch: the auto-upload runner now handles the tiered gate per trace, and the docs describe the new invariant. This is the PR that turns "33 candidates → 17 disappeared" into "most ship, a few auto-redacted, one or two pending review".

## Runner changes (`_run_cycle_impl`)

- **Mapped block/review findings**: only the affected traces move to `pending_review`; the clean remainder is **repackaged once in the same cycle** (bounded single retry, mirroring the manual CLI's remove-and-retry loop). All downstream bindings — `selected`, `expected_revisions`, raw fingerprints, strict-scan binding — are narrowed to the survivors and the control-state assertions repeat before the second packaging attempt, preserving the "controls win before submitting" and exact-bytes seal invariants (the seal ledger only ever sees the narrowed set).
- **Unmappable findings**: the whole batch parks in `pending_review` with an explanatory hold reason and the cycle returns retryable `review_attention`. Previously this raised the durable, non-retryable `unmappable_findings` — health went `action_required`, every future cycle stalled, and the sessions stayed candidates re-failing forever. A human now un-sticks it from the workbench; nothing is lost.
- **Scanner unavailable** (`scanner-not-installed` / `scanner-error`): retryable `scanner_unavailable` with zero hold-state changes — infrastructure trouble never parks a trace.
- Redact-tier findings need no runner change at all: the gate rewrites inside `package()` and those sessions ship in the same batch.

## Docs

- **CLAUDE.md**: the "Mandatory TruffleHog post-redaction gate" invariant is now "Mandatory tiered secret-scan gate" (Betterleaks primary, validation never enabled; TruffleHog verified-only, AGPL subprocess-only; tier semantics; fail-closed manual / retryable auto; redaction completes before hash/seal; bypass recorded + refused on upload). Module map gains `betterleaks.py`/`scan_policy.py`/`share_gate.py`.
- **PRIVACY.md**: tier policy table-in-prose, the no-candidate-secret-egress statement (detection layer never calls provider APIs), both managed installs, new `secret-scan*.json` artifacts, updated bypass semantics.
- **README.md**: both scanners in the Stage-6 install block and command table; auto-upload section notes per-trace review routing.

## Tests

Rewrote `test_unmappable…requires_action_instead_of_retry_loop` → `test_unmappable_finding_parks_batch_for_review_instead_of_stalling` (asserts healthy enrollment + per-session `pending_review` + retryable result), added `test_mapped_findings_park_only_bad_traces_and_retry_batch_same_cycle` (asserts exactly two packaging attempts with narrowed batch) and `test_scanner_unavailable_is_retryable_without_hold_changes`. Full suite: **2911 passed**.


kai-rayward:
**Adversarial review: merge-ready.** All eight invariants traced and confirmed by an independent reviewer: the retry loop is bounded to exactly two package attempts, hold/revision/fingerprint bindings narrow to survivors before the retry, `_assert_control_state` re-runs between attempts, the sealed ledger references the retry's artifact, empty-survivor concludes cleanly, and `scanner_unavailable` touches no hold state. Landed since: `test_mapped_finding_retry_ships_survivors_with_narrowed_scope` — the headline happy path (attempt 1 parks one trace, attempt 2 seals and submits the survivor) asserting the narrowed revisions, fingerprints, and sealed-ledger scope end-to-end.

Known minor gaps left as-is (non-blocking): if the *second* attempt raises (e.g. transient scanner trouble), the cycle result carries that error code without a count of the traces already parked by attempt 1 (state stays coherent and audit-logged in hold history); a `secret-scan-bypassed` block on the auto path maps to retryable `package_failed` rather than a distinguished outcome (unreachable in production; it still never ships).

PR #135, opened 2026-07-22
Enhance submission success celebration

richardopenrefinery:
## Summary

* replace the upward completion burst with four stronger top-down confetti and light waves spanning the full success sound
* play the supplied 9.311-second MP3 after a successful hosted submission
* arm the reusable audio element silently from the original Submit click so Safari preserves playback permission across the asynchronous upload
* stop and reset silent priming on expired verification and failed submission paths
* preserve reduced-motion behavior and add focused tests for exact-asset playback, delayed success, and failure cleanup

## Why

The existing submission confirmation effect burst upward from the middle and was visually understated. The updated flow makes the completion moment more celebratory, sends the effect down from the top, and synchronizes repeated waves with the approved sound.

The first implementation only loaded the MP3 during the Submit click and called `play()` after asynchronous validation and upload requests. Safari requires playback to originate from the user gesture, so it could reject the delayed call. The final implementation starts the same media element silently during the click, then restarts it audibly only after hosted submission succeeds.

## User impact

After a successful hosted submission, users see a denser top-down celebration and hear the approved success recording once, including in browsers with direct user-gesture playback rules. Failed submissions and local-only bundle creation remain silent.

## Validation

* `npm test` — 42 tests passed
* `npm run build`
* targeted ESLint on the changed Share UI and sound files
* `git diff --check`
* source, public asset, and built asset SHA-256 verified identical: `CCBFCC5642675F8A148F14D9714431F533E758F27DA778310D83B7ADA7AB9705`


richardopenrefinery:
I tested it today and noticed that the celebration animation—the upward burst with confetti and light waves—was removed, leaving only the sound?

PR #144, opened 2026-07-23
feat(auto-upload): streamline recurring enrollment and remove the UI rollout gate

JustMeStudio:
## Summary

- Cache the short-lived recurring-enrollment grant returned after a successful hosted manual share.
- Use that grant to create protocol-v2 recurring enrollment without asking the participant to verify the same email a second time.
- Present the exact recurring authorization, retention terms, scope, and distinct ownership certification directly on the successful receipt flow.
- Simplify CLI enrollment retries while preserving exact-version and post-refresh scope acceptance.
- Remove the internal `CLAWJOURNAL_ENABLE_AUTO_UPLOAD_UI` rollout gate so Automatic uploads is visible by default, and migrate away the retired config key.

## Safety invariants

- Recurring authorization remains separate from manual-share consent.
- Enrollment remains protocol-v2, future-only, exact-scope, ownership-certified, capped at five traces, and fail-closed on scope-hash drift.
- The receipt-issued grant is one-time, origin-bound, expiry-checked, removed after use, and falls back to normal email verification when unavailable or rejected.
- Manual Share review, redaction, secret scanning, and submission behavior remain unchanged.

## Validation

- Full Python suite: **3028 passed, 4 skipped**.
- Focused grant, client, config, CLI, and daemon suites: **215 passed**.
- Frontend tests: **50 passed**; production build completed successfully.
- Regression coverage includes ISO and epoch grant expirations, rejected-grant email fallback, strict integer capability versions, origin binding, credential cleanup, and conditional receipt copy.
- Local Docker E2E: manual hosted share → enrollment grant cached → recurring enrollment enabled without repeat email verification → post-enrollment trace selected → recurring artifact uploaded → matching receipt persisted by the hosted service.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -31,7 +31,7 @@ tests/
 - `clawjournal/export/`: renders export formats such as JSONL and Markdown.
 - `clawjournal/prompts/`: canonical runtime prompt assets used by scoring and PII review.
 - `clawjournal/auto_upload.py`: SQLite-owned recurring authorization state, candidate selection, cadence, exact-artifact sealing, crash recovery, and the fail-closed runner.
-- `clawjournal/auto_upload_client.py` and `auto_upload_credentials.py`: the typed hosted v1 protocol and purpose-separated private credential boundary.
+- `clawjournal/auto_upload_client.py` and `auto_upload_credentials.py`: the typed hosted v2 protocol and purpose-separated private credential boundary.
 - `clawjournal/agent_hooks.py`: semantics-preserving Claude Code/Codex `SessionStart` configuration plus the bounded due-check adapter.
 
 ## Data Flow
@@ -82,6 +82,7 @@ Optional self-hosted path:
 Optional hosted recurring path:
 
 - It remains unavailable unless hosted discovery advertises protocol v2 and the participant has a successful manual receipt.
+- A successful manual share may return a short-lived, single-use enrollment grant bound to that receipt and participant. The client prefers that grant for immediate enrollment and falls back to fresh email verification when it is unavailable, expired, or rejected.
 - Enrollment sends the explicit (source, project) scope entries and requires two distinct affirmative acts: the versioned recurring authorization/retention acceptance and the versioned ownership certification. The server computes and owns the scope hash; the client pins the value read back at enrollment and fails closed on any drift.
 - The local SQLite singleton owns mode, generation, accepted exact scope/profile, cadence, health, and run overlay. Private files own active/recovery credentials; `config.json` never does.
 - The hosted service owns versioned recurring authorization, credential hashes, exact-byte idempotency, cross-enrollment duplicate-revision rejection, storage, and receipts.
--- a/clawjournal/auto_upload.py
+++ b/clawjournal/auto_upload.py
@@ -44,6 +44,7 @@ from .auto_upload_client import (
     fetch_authorization,
     fetch_capabilities,
     get_enrollment,
+    grant_capability_version_supported,
     lookup_receipt,
     recovery_capabilities,
     revoke_enrollment,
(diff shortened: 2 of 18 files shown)


kai-rayward:
Reviewed this and pushed a follow-up commit (938e4c9) rather than sending it back — all five checks are green on the result. Retitled the PR too, since the body already covered the rollout-gate removal but the title didn't.

The first item below is a judgment call, and I'd genuinely like you to push back if you disagree with it.

**Restored the typed-version consent ceremony in `_interactive_accept`** (plus the README sentence and the tests). This reverts the `[y/N]` prompts to typing the exact authorization, retention, and ownership versions. The reasoning: transcribing the version *is* the affirmative act — a bare y/N can be answered without reading the text it's attached to, and it turns the accepted version into a value the client echoes back from the challenge rather than one the participant attested to. Recurring sharing uploads future traces with no per-bundle review, so it keeps the heavier gate even though the web checkboxes are lighter.

Nothing else about the enrollment flow changed. Enabling straight from a receipt still skips the second email code, which is the actual friction this PR set out to remove — and with the versions typed again, the "preserving exact-version acceptance" bullet in your description is now true of the UX as well as the protocol.

**Fixed a silent origin-comparison bug in the grant path.** `_hosted_api_base()` echoes whatever casing the operator wrote in `CLAWJOURNAL_SHARE_URL`, but it was compared against `capabilities["origin"]`, which lowercases the host. Any uppercase character in the configured host made *every* grant permanently unusable, falling back to email verification forever with no error surfaced — the feature would just have looked broken on those installs. Both sides now normalize through a shared `comparable_origin()`, and the issuer is stored normalized.

Worth noting this normalizes spelling, not authority: a differing port is still a differing origin. My first attempt at the test asserted `:443` should equal the bare host, which would have quietly loosened the exact-origin pinning the rest of the module relies on — there's now an `:8443` case locking the strict behavior in.

**Bound `status()`'s `enrollment_grant_available` to that same origin check.** The workbench reads the field to decide whether it may skip email verification, so computing it from local presence and expiry alone pointed the participant down a path `enable` was bound to refuse. An install with no resolvable hosted origin now reports no grant. The hosted capability version is deliberately still not checked there — that needs discovery, and `enable` stays the authority on it.

**Renamed `create_identity_kind` to `sent_identity_kind`.** It's also set on the credential-rotation path, not just create, and the cleanup keyed off it deletes a still-valid credential if it's read as create-only. I traced that path carefully and the logic was correct — this is purely to stop the name misleading the next reader. Also added a comment marking the `"auto"` branch of `_hook_targets` as a deliberate guard: it's unreachable from `enable` today because the scope blockers already reject anything outside `COMPLETION_MODES`.

Everything else held up under review: the manual-receipt gate, the ownership certification, scope-hash pinning, the five-trace cap, and default-off are all intact, and the grant is properly one-time, expiry-checked, and cleaned up on every failure branch I traced. The XOR on `create_enrollment`'s two identity credentials is a good addition — it makes "exactly one verified identity" structural rather than conventional.

kai-rayward:
**Update: this PR is now standalone and merges FIRST** (order reversed from the earlier note — #148 now stacks on top of this branch).

The branch is Dean's five commits rebased onto current main, plus one review-fix commit. No #148 content is included.

What changed relative to the original commits:

- **Visibility policy**: the conflict with #147's conditional `ui_visible` was resolved in favor of this PR's always-visible model (Kai's call). #147's conditional-visibility test became an `offer_available` test — the stale-config-cache protection it locked in still holds, at the offer level.
- **README**: the paragraph edit targeted the pre-#145 README, which no longer exists; ARCHITECTURE.md carries the doc update.
- **Grant handling (review fixes, `d745891`)**: the receipt grant is now preferred over a live `verified_email_token` (consuming the manual-share token while the grant idles would cost the next manual share); an unknown advertised grant version now only disables the grant path instead of failing the whole capability document (exact-integer check moved to `grant_capability_version_supported`, bools still rejected); grant config keys deduplicated into `config.py`; new tests for the preference order, version degradation, and identity-switch grant clearing.

Validation on this branch: full Python suite **3046 passed**, frontend **66 passed**, targeted ESLint clean, production build green.

PR #148, opened 2026-07-23
Guide Auto Upload scope setup and contain success confetti

timyeyuqing:
## What changed

- Turn source/project scope blockers into an inline setup step inside the Auto Upload authorization dialog.
- Limit the guided recurring-upload scope to the sources the unattended runner currently audits: Claude Code, Codex, or both.
- Route an existing `unsupported_source` configuration back into that guided setup instead of leaving the user at another dead end.
- Mark projects with the backend's effective exclusion policy and omit excluded projects from the confirmation list.
- Add an explicit **Confirm all eligible projects** action, including the exact count being confirmed, before the scope can be saved.
- Require a fresh project confirmation each time guided setup opens and whenever the source selection changes.
- Hide the separate run-trigger picker during scope setup so two similarly labelled source controls cannot be confused.
- Save source and project confirmation together, then continue automatically to the existing exact-scope, recurring-terms, ownership, and email-verification steps.
- Keep Settings synchronized even if the authorization dialog closes while the scope write finishes.
- Replace the CLI-oriented oversized-scope error with actionable UI guidance.
- Constrain the submitted-state flash and confetti to the Share main-content container, with a top-down fall that no longer covers the sidebar.

## Why

A fresh install could reach Auto Upload with no confirmed export scope, or with a global `all` scope that included agents the unattended runner cannot safely snapshot. The UI exposed every manual-export source and listed raw indexed projects, while recurring enrollment supports only Claude Code and Codex and applies project exclusions before authorization. Those mismatches could either dead-end the enable flow or show a project list different from the exact authorized scope.

The success celebration also used viewport-fixed positioning, so it covered the navigation sidebar instead of staying inside the Share surface.

## User impact

After one successful manual share, users can repair missing or unsupported scope configuration without leaving the Auto Upload dialog. They explicitly confirm all currently eligible projects, authorize only supported recurring sources, and see only projects eligible after effective exclusions. The success effect now drops through the Share content while leaving navigation unobstructed. Manual Share behavior and every existing backend privacy, redaction, hold-state, exact-scope, ownership, and email-verification gate remain unchanged.

## Validation

- `npm test -- --run` — 71 tests passed
- `npm run build` — production TypeScript/Vite build passed
- Targeted ESLint on every changed frontend source/test file passed
- `pytest tests/workbench/test_daemon_config.py -q` — 17 tests passed
- Python suite — 2,990 tests passed; the 28 trace-note tests affected by a local third-party `tests` package collision also passed in an isolated import run
- Scope-save failure regression test verifies that the dialog, confirmation, and retry action remain available when persistence cannot be confirmed
- Success-effect regression test verifies that all three flash layers and the confetti overlay are absolutely positioned inside the Share shell
- Oversized-scope regression test verifies that the guided UI returns actionable guidance instead of a CLI-only error
- Visual QA of the real `DoneStep` component at 1280px: sidebar occupied `0–190px`; confetti and Share shell both occupied `190–1280px`, with no overlap
- Interaction QA against an isolated unsupported-source fixture: unsupported `all` scope → choose Claude/Codex → excluded projects omitted → confirm all eligible projects → save → exact recurring authorization
- `git diff --check`

The Vite build retains the repository's existing bundle-size warning. The repository-wide lint command also retains unrelated pre-existing errors outside this PR's changed files.

### DIFF
--- a/clawjournal/config.py
+++ b/clawjournal/config.py
@@ -198,9 +198,10 @@ def set_source_scope(config: ClawJournalConfig, source: str) -> None:
 def source_scope_sources(source: str | None) -> tuple[str, ...] | None:
     """Return allowed session sources for a confirmed source scope.
 
-    ``None`` means unrestricted/all sources. Legacy ``both`` means the
-    original Claude+Codex pair, while ``all`` intentionally means every
-    supported indexed source.
+    ``None`` means unrestricted/all sources. ``both`` means the Claude+Codex
+    pair — it predates ``all`` and is also what the Auto Upload guided scope
+    setup writes, since those are exactly the recurring-capable sources —
+    while ``all`` intentionally means every supported indexed source.
     """
     normalized = (source or "").strip().lower()
     if normalized in ("", "auto", "all"):
--- a/clawjournal/web/frontend/src/components/AutoUploadControls.test.tsx
+++ b/clawjournal/web/frontend/src/components/AutoUploadControls.test.tsx
@@ -2,7 +2,7 @@ import { act, fireEvent, render, screen, waitFor, within } from '@testing-librar
 import { MemoryRouter } from 'react-router-dom';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import { api, ApiError } from '../api.ts';
-import type { AutoUploadStatus } from '../types.ts';
+import type { AutoUploadStatus, WorkbenchConfig } from '../types.ts';
 import { AutoUploadOffer, AutoUploadPanel } from './AutoUploadControls.tsx';
 import { ToastProvider } from './Toast.tsx';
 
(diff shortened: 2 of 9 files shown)


kai-rayward:
**Now stacked on #144** — merge order is **#144 first, then this PR**.

⚠️ This branch temporarily contains #144's commits, so the diff here includes them until #144 merges. This PR's own content is the single top commit `c672c37`, which also folds in the review fixes previously pushed as `4b4d412` plus the reconciliation with #144's inline receipt offer:

- The guided scope step wins the dialog copy/controls in every inline/modal state; the "Run on agent sessions" select hides during scope setup (its option labels duplicated the source-scope select).
- A single-source scope save retries with the matching agent hook — a claude-only scope never installs a codex SessionStart hook.
- The offer tests run against #144's auto-open inline flow; the consent checkbox never pre-checks from stored config; `scope_too_large` gets a UI-worded error.

After #144 squash-merges, re-anchor with:

```
git fetch origin main
git rebase --onto origin/main d745891 codex/inline-auto-upload-scope
```

(`d745891` = the last #144 commit on this branch.)

Validation on the stacked tip (byte-identical to the previously validated combined tree): full Python suite **3047 passed**, frontend **74 passed**, ESLint clean, production build green.

PR #151, opened 2026-07-24
fix(selfupdate): keep frontend and daemon versions aligned

JustMeStudio:
## Summary

- capture an immutable frontend snapshot before the detached self-updater can move the checkout or rebuild `dist/`
- keep each running workbench daemon on the frontend bundle that matches its imported Python, then switch the frontend/backend pair together on the existing graceful restart
- preserve disk-backed frontend behavior for `--reload` and direct development server callers
- make the updater lock test cross-platform, add a real Windows PowerShell reinstall subprocess test, and add a focused `windows-latest` CI job
- clarify that explicit `selfupdate --reinstall` intentionally reruns the installer to repair unrecorded environment drift

## Why

PR #149 rebuilds the gitignored frontend in the background while an already-running daemon continues serving files from `dist/`. That creates a window where a browser can receive newly built frontend assets from an old Python backend. Pinning a process-local snapshot removes that mixed-version window without changing the ff-only update, pending-reinstall, or graceful-restart safety gates.

## Validation

- `python -m pytest tests/test_selfupdate.py tests/test_install_scripts.py tests/workbench/test_update_restart.py -q -p no:cacheprovider` — 90 passed, 19 skipped on Windows
- focused frontend snapshot, static serving, restart ordering, lock, and desktop handoff tests — 23 passed
- static serving and desktop launch regression selection — 7 passed
- `git diff --check`


kai-rayward:
## Review

The diagnosis here is right and the fix is at the right layer — pinning in memory rather than trying to make `dist/` atomic on disk is the pragmatic call, and at 536K / 5 files the memory cost is nothing. Nice catch on the `reinstall_needed` docstring too: its only non-test caller is the restart gate at `daemon.py:5616`, so the old "safely run unconditionally / the installer run is skipped" claim described an optimization that was never actually wired to `reinstall()`.

I found one bug that I think would have bitten in normal operation, so I pushed a fix plus regression tests directly to the branch (`ae2cb30`), and then a second commit for the smaller items (`db5c220`). Details below — happy to revert either if you'd rather take a different approach.

---

### The bug: an empty snapshot is fail-*broken*

`capture_frontend_snapshot` returned an empty `FrontendSnapshot` on every failure path, and `_serve_static` treats "snapshot present, no `index.html`" as *serve the placeholder* — with no reason to ever re-read disk again. So an empty pin blanked the workbench for the entire life of the process.

That's reachable in normal operation. `vite.config.ts` sets no `emptyOutDir`, so Vite's default (`true`) applies: `npm run build` **clears `dist/` first**, then bundles for seconds before writing anything. The installer runs that build (`scripts/install.sh:251`). So a `serve` / `desktop launch` that starts while an updater spawned by *an earlier* CLI invocation is mid-build samples a stable-but-empty tree — the before/after consistency check passes trivially, because `()` equals `()`.

Reproduced both halves:

```
# capture, during the bundling window
snapshot during vite bundling window -> files: 0  index.html: None
dist on disk afterwards: ['index.html', 'new.js']      # dist is healthy

# serving, with that snapshot pinned and dist/ healthy on disk
GET /                  -> 200 b'<!DOCTYPE html>... ClawJournal Workbench'   (placeholder)
GET /assets/app.js     -> 200 b'<!DOCTYPE html>...'                          (placeholder, as text/html)
```

Pre-PR this self-healed on the next request, so it's a regression rather than a pre-existing edge.

**Fix (`ae2cb30`):** return `None` instead. The caller contract already handled it — `cli.main`'s `except Exception: pass` and `_serve_static`'s `snapshot is None` branch both already mean "serve off disk", so no caller changes were needed. Alongside it: require `index.html` before pinning (a half-written tree has no SPA entry point), skip the retry loop entirely when `dist/` doesn't exist, and put a 50 ms gap between attempts — back to back they re-sampled the same instant, so the retry only ever caught a read that lost by microseconds.

Five regression tests, four of which fail on a real assertion against the unfixed code (`assert FrontendSnapshot(files=mappingproxy({})) is None`) rather than just on the new kwarg. The daemon-level one is the end-to-end proof: capture during an emptied `dist/`, let the build land, confirm the server returns the real bundle instead of the placeholder.

### Smaller items (`db5c220`)

- **`_warn_if_frontend_stale` described the pre-snapshot world** — "the editable install serves the gitignored `dist/` straight off disk" — and its advice was wrong for a pinned process, since rebuilding can't reach it. It now takes `pinned=` : the STALE hint asks for a restart too, and a `dist/` the installer has since emptied no longer draws a MISSING banner (the pinned copy is what's being served).
- **Scoped the pin to where it can help.** A wheel install has no checkout for the updater to fast-forward, and `desktop status` / `desktop stop` exit without ever binding a socket — both were reading `dist/` into memory for nothing. New `_should_pin_frontend` collects that with the existing `--reload` opt-out.
- **Documented the dev consequence** in `CLAUDE.md` + `AGENTS.md`: a running `serve` won't show a rebuild until it restarts (`serve --reload` still will). Worth saying out loud since the frontend section otherwise just says "run `npm run build`".
- **`_tree_signature`** — one `stat` per file via `os.walk` instead of two via `rglob` + `is_file()`. Verified the file set is byte-identical including dotfiles.
- **`tests/test_install_scripts.py`** reported a *pass* while silently dropping its subprocess assertion when `sh` was missing; `pytest.skip` instead. (Turns out Windows CI does find Git-Bash's `sh`, so that assertion really runs there — this is a guard, not a hole being papered over.)

One judgment call worth flagging: the `os.walk` rewrite is slightly stricter than the original — a file that can't be `stat`'d now aborts the signature (→ retry, then disk fallback) where `is_file()` silently omitted it. That's the conservative direction and consistent with the module's "any doubt, don't pin" stance, but it is a semantic change rather than a pure optimization.

### Verified as correct

- Ordering: capture strictly precedes `maybe_self_update()`, and the test asserts it.
- `_exec_restart` re-execs through `clawjournal.cli`, so the new process re-captures — and it only fires once `reinstall_needed()` is false, so the primary update path never races its own build.
- `--reload` is genuinely preserved: `_reload_child_command()` keeps `--reload` in the child argv, so the child stays disk-backed.
- Path traversal: snapshot keys come from `relative_to(root).as_posix()`, so nothing outside `dist/` is reachable, and the 403 check still runs first — it normalizes `..` lexically even when `dist/` has been deleted.
- Snapshot reaches both sockets, including the port-fallback branch and the IPv6 companion.
- Content-type behavior is equivalent; `Path("assets/app.js").suffix` works on Windows too.
- Immutability: `MappingProxyType` over `bytes`, and `_inject_api_token` returns new bytes rather than mutating.

### Follow-up, not for this PR

The guarantee is per-*process*. After the re-exec, an already-open tab still holds the old bundle against a new backend — there's no version poll or reload prompt in the frontend (only `ErrorBoundary`'s manual reload button). This PR closes the strictly worse case (a *fresh* page load getting new assets from an old backend); a "new version available — reload" nudge would close the rest.

---

Full suite on the branch: **3146 passed, 1 skipped**. CI green on `db5c220` across `test (3.10–3.13)`, `windows-selfupdate`, and `smoke` — and the new `windows-selfupdate` job is a real win, since `test_windows_reinstall_targets_the_active_interpreter` had never actually executed anywhere before this PR.

PR #153, opened 2026-07-25
Add evidence-backed weekly focus

kai-rayward:
## Summary

- Add a deterministic `Focus this week` spotlight to `clawjournal skill --preview`.
- Select one freshly distilled avoid rule only when its current-run evidence resolves to at least 3 real sessions across 2 days and 2 canonical projects.
- Require every counted session to contain the rule's valid failure taxonomy; a do-pool session must also record a clean recovery.
- Block unsupported personal/work-performance claims from the proposed and installable rule set, not only from the spotlight.
- Harden semantic dedup so title similarity requires independent meaning corroboration and cannot erase parallel or extended scoped lessons.
- Update the shipped ClawJournal skill trigger, README, and self-improving-skills design documentation.

## Why

The proposed critique prompt has a useful shape—a single behavior, its downstream cost, and one replacement habit—but its original wording overclaims what coding-session traces can establish about a person or their performance at work. Existing stored `case-NN` aliases and taxonomy-wide support counts are also not exact long-lived provenance.

This implementation keeps the useful prioritization structure while grounding it only in fresh current-run citations. Stored aliases, synthetic aggregate candidates, ambiguous duplicate aliases, source-prefix project inflation, carried paraphrases, unrelated successes, and personal/work-performance diagnoses cannot qualify for the spotlight or enter newly rendered agent instructions.

## User impact

`clawjournal skill --preview` now shows either:

- one coding-agent behavior with its observed cost, replacement trigger, replacement habit, and evidence breadth; or
- an explicit abstention.

The spotlight adds no second model call. Its framing and aggregate evidence remain preview-only, while the underlying lesson stays in the proposed skill set and follows the existing confirmation/install flow.

The shipped `clawjournal` skill also recognizes natural requests such as “improve my Codex usage” and uses already-scored history before asking permission for broad historical AI scoring.

## Review follow-ups

- `c291109` corrected abstention reasons and extended the personal-claim safeguard to the displayed replacement trigger.
- `afb94e1` made cleanly recovered failures eligible even though candidate selection routes them into the do pool.
- `4c8c191` added near-title paraphrase dedup for the observed “Prove/Reproduce Review Findings” duplicate.
- `d44ba3a` closes the initial review gaps:
  - evidence requires the same structured failure taxonomy, plus a canonical clean-recovery label for do-pool candidates, so unrelated successes cannot inflate the spotlight;
  - personal/work-performance diagnosis checks cover every displayed field and common possessive, copular, adverbial, role-bound, and procrastination forms while preserving technical wording such as an unreliable test;
  - fuzzy title dedup requires matching taxonomy or corroborating guidance/trigger vocabulary;
  - the distill prompt frames `why` as a task, artifact, or coding-agent consequence rather than a personal trait or overall-work assessment.
- `f5c95a9` prevents weak fuzzy-title evidence from collapsing parallel scoped lessons:
  - equal title templates with one non-leading scope substitution, such as API versus database responses, fall through to the stronger full-guidance overlap check;
  - the ordered title tokenizer retains short scope terms such as `DB` and `AI`;
  - action-word rewrites and strongly corroborated paraphrases still collapse.
- `648c91c` applies the remaining policy fixes:
  - a shared deterministic rule-policy gate removes unsupported personal claims before merge and checks the full merged set again before rendering, covering fresh, fingerprint-colliding, and mixed legacy rules;
  - technical phrases such as “avoid unreliable integration tests” remain allowed;
  - title extensions such as `Test AI Security` require the same independent taxonomy/guidance/trigger corroboration as other fuzzy matches, while the measured `Fix Root Cause Durably` paraphrase and strong full-guidance matches still collapse.
- `97cd003` hardens the personal-evaluation boundary:
  - structural human-subject checks reject unsupported judgments across pronouns and human roles instead of depending on a narrow trait list;
  - ownership-aware matching blocks human evaluations such as reviewer competence and poor ownership while allowing technical context such as model decisions, reviewer judgment records, component ownership, authorization, and artifact quality;
  - direct work-performance forms and coordinated attempts to hide an evaluation behind technical wording are rejected;
  - focused regression matrices cover both the deny boundary and common technical false-positive cases.
- `49452fc` removes the technical-noun escape identified on the previous head:
  - arbitrary human predicates such as “inept at code review” are rejected even when they mention a technical noun;
  - only explicit operational predicates can establish safety;
  - later coordinated technical objects may inherit a recognized action through a bounded technical noun phrase, preserving wording such as “fixing code and unit tests” without reopening the escape;
  - common coding actions and technical ownership roles remain accepted.
- `b046854` closes the two follow-ups from the prior head review:
  - sequence-aware title comparison detects a non-leading scope substitution even when one title adds a modifier, so parallel API/database lessons survive while leading action rewrites and pure extensions still deduplicate;
  - a bounded morphological fallback accepts unlisted gerunds only when they directly govern a technical noun phrase;
  - state and performance gerunds remain denied, including “being inept,” “underperforming on tests,” and “performing poorly in code review.”
- `c1703f4` applies the final adversarial hardening:
  - stored legacy rules are policy-gated before semantic clustering, so unsafe carried wording cannot suppress a safe fresh replacement and then disappear;
  - main-clause manner adverbs and evaluative decision phrases are denied structurally across linking and ordinary action verbs;
  - bounded technical actions such as migrating a production database, bootstrapping a fresh database, and traversing a dependency graph remain allowed;
  - technical plural antecedents are distinguished from human `they` in linking clauses.

- `dfcb994` preserves valid operational lessons found during live dogfooding:
  - `reply` is no longer misread as a manner adverb, and narrowly grounded `correctly diagnosed` / `fix it properly` evidence remains eligible;
  - completed investigation wording is recognized as technical context;
  - exact-token regression tests ensure unrelated unsafe adverbs cannot inherit either exemption.

- `58d9902` closes the current-head policy and preview findings:
  - benign task adverbs such as `carefully reviewing`, `proactively testing`, and `systematically validating` remain eligible while explicit evaluative manner claims stay denied;
  - subordinate human-evaluation clauses are checked without attributing later technical-subject predicates to the person, including punctuation, agreement, and `they` antecedent cases;
  - an all-blocked preview now prints each dropped rule and its safety-gate reason instead of returning before those diagnostics.

## Validation

- `PYTHONPATH=. pytest -q` — **3,540 passed, 1 skipped**
- `PYTHONPATH=. pytest -q tests/skill` — **542 passed**
- focused policy/preview/generation/focus regression suite — **414 passed**
- `PYTHONPATH=. python -m clawjournal.cli skill --no-scan --no-score --preview` — five-rule preview generated; weekly focus correctly abstained; nothing installed
- `python -m py_compile clawjournal/cli_skill.py clawjournal/skill/rule_policy.py` — passed
- `git diff --check` — passed

The full-suite run redirected ClawJournal's index and blob paths to a fresh temporary directory, so verification did not touch the user's real local database. No frontend files changed, so a frontend rebuild was not required.


kai-rayward:
## Review of the policy/focus hardening — 2 fixed in `df472af`, 9 still open

I ran an adversarially-verified review over `d44ba3a..dfcb994` plus `rule_policy.py` (5 independent lenses, every finding independently refuted before counting; 22 raised → 16 confirmed). I then **re-verified every item against the current head** before posting, so the open list below excludes anything already fixed.

### Fixed in `df472af`

**Evidence days were bucketed by UTC.** Sessions are stored as UTC instants, so west of UTC one unbroken sitting that crosses UTC midnight was reported as two days — three sessions at 15:40 / 16:20 / 17:10 US-Pacific printed `across 2 days` for a 90-minute block. That is exactly the one-sitting case the `>= 2 days` threshold exists to reject; 70 such clusters exist in one real 928-session index. Now counted in local days, matching the convention `desktop.py` already uses for its user-facing day count.

**The actor vocabulary was a closed list.** `"the user is careless"` was blocked while every one of these passed: `"the human is careless"` (and *human* is the likeliest word for a model summarizing agent transcripts), `"the dev is sloppy"`, `"the teammate is unreliable"`, `"whoever wrote this was negligent"`.

`dev`/`folks` join only the clause-level actor alternation (which requires a following copula) and deliberately stay **out** of `_PERSON_ACTOR_NOUN`, since that constant also feeds the attributive check where they would reject `"unreliable dev server"`. Accepted residual, documented in the commit: a second person introduced as "dev" in a multi-sentence plural antecedent (`"The human reviewed tests with the dev. They were inept."`) still passes. Closing it reinstates the false positive, and silently deleting a real lesson is the worse failure.

### Still open — false positives (verified at `df472af`)

These destroy legitimate lessons. Since the gate now runs over carried rules too, they can silently remove previously-installed lessons on upgrade.

| wording | field | verdict |
|---|---|---|
| `when you are unsure, ask before proceeding` | guidance | blocked |
| `clear the cache when your cache is stale` | guidance | blocked |
| `when you are about to rename a symbol, search for callers` | guidance | blocked |
| `unreliable user tests masked the failure` | why | blocked |
| `you reported the badge incorrectly` | why | blocked |

Root causes the verifiers isolated: `_is_safe_link_predicate` uses a closed safe-set (`_SAFE_RESOURCE_STATE_RE` has `uncertain` but not `unsure`); `_possessed_subject_is_technical` requires the possessed noun in a closed technical list with no `cache`/`endpoint`/`flag`/`symbol`/`module`; `_PERSONAL_PRENOMINAL_RE` doesn't check that the role noun is the phrase head, so it reads `unreliable [user tests]` as `unreliable user`; and the `_SAFE_ACTOR_ADVERBS` allowlist contains no adverb of correctness or timing, which is the literal content of an `avoid` rule's `why`.

### Still open — false negatives (verified at `df472af`)

| wording | verdict |
|---|---|
| `the user in this session is careless` | passes |
| `the developer, reviewing quickly, is sloppy` | passes |
| `the developer's inability to read a diff caused the regression` | passes |
| `the user's unwillingness to run tests slowed the fix` | passes |

The first two: any word between the actor and the copula defeats the adjacency requirement (`_PERSON_LINK_AFTER_MODIFIER_RE` covers subordinators like `who`/`which`/`when`, but not a bare prepositional phrase or appositive). The last two: `_possessed_subject_is_technical` accepts a possessed subject whenever a technical noun appears *anywhere* in it, so a nominalized personal deficit owned by the human is laundered by its technical object.

### Already fixed upstream — no action needed

The `-ly` false positive (`apply`/`rely`/`comply`), `"the agent unnecessarily re-ran the suite"`, `_same_lesson` asymmetry, and the preview hiding policy-gate drops all re-tested clean at the current head.

### Related

Separately filed #156: success candidates crowd failures out of the 15-slot distill pool (14 of 15 slots go to successes whose `support` comes from `outcome_counter[badge]` rather than a failure mode), so no real failure candidate reaches the distiller and the spotlight abstains on a real corpus. **That is not a defect in this PR** — abstaining is correct when no citable evidence arrives — but it means the feature ships dormant until selection is fixed.

Full suite at `df472af`: 3,556 passed, 1 skipped.

kai-rayward:
## Correction + the false positives are fixed (`c520da9`)

**Correction to my earlier comment.** I listed `you reported the badge incorrectly` (in `why`) as a false positive. It is not — `why` bans second person by design, and the adverb is irrelevant: `why="you reported the badge"` is equally blocked, while `why="the badge was reported incorrectly"` passes. Apologies; four of the five were real.

### Fixed

`when you are unsure, ask before proceeding` · `when you are about to rename a symbol, search for callers` · `clear the cache when your cache is stale` · `unreliable user tests masked the failure`

Causes were: an intransitive state was only safe beside a technical noun, and *unsure* never has one; the linking patterns capture to the end of the sentence, so the main clause after a subordinate comma was judged as part of the complement; `cache`/`symbol`/`module` and ~40 other everyday nouns were missing; and the attributive check never asked whether the role noun was the head, so `unreliable [user tests]` read as `unreliable user`.

### The part worth reviewing

Loosening a deny gate is how you open one, so I attacked each change adversarially. **It found 17 new bypasses** — the first cut traded 4 false positives for 17 false negatives, which is a far worse deal. All 17 are now closed and kept as regression tests:

- **Head-noun matching.** `the user's memory of the folder layout is unreliable` passed as soon as `folder` became technical — the possessed-subject and attribution tests searched the *whole* phrase, so a trailing prepositional phrase laundered a personal head noun. They now match on the head, before any preposition. This also closes pre-existing siblings like `the developer's inability to read a diff`.
- **Standing dispositions.** `has a history/habit/pattern of ...` stays personal however technical the rest is. The coding agent is exempt — describing what it repeatedly does is the point of an avoid rule — unless trait words appear.
- **Clause truncation** now applies only to a genuinely new clause. A coordinated continuation (`..., and has a long history of approving diffs without reading them`) or any remainder carrying trait vocabulary keeps the full predicate.
- **Bare states** are safe only for `you`/`we`/the agent; `the reviewer is unsure about nearly every change` evaluates a third party and still needs an object.
- **The attributive exemption** is limited to `unreliable` + `user`/`agent` + a closed list of nominal heads, so `an unreliable user commits secrets` and `careless developer commits` stay denied — a general "any technical word follows" test also matches verbs.

### Verification

Full suite **3,593 passed, 1 skipped**; the new deny tests fail against the pre-change gate. Checked against the 25 real stored rules in a live index: **0 changed verdict**, so no live lesson is gained or lost by this change.

One unrelated flake seen once in `tests/workbench/test_daemon.py::test_finalize_reraises_control_gate_instead_of_swallowing[control_changed]`; it passes in isolation both with and without these changes, and passed in five prior full runs.

### Still open from the earlier review

The false negatives are unchanged and still worth a look: `the user in this session is careless` and `the developer, reviewing quickly, is sloppy` — any word between the actor and the copula defeats the adjacency requirement.

PR #162, opened 2026-07-27
fix(share): reconcile continued traces before hosted upload

JustMeStudio:
## Summary

- use the hosted service's advertised manual-lineage preflight before manual upload
- rebind an unsubmitted share to the receiver-authoritative predecessor when the participant created the reviewed share after that hosted head was accepted
- invalidate any finalized ZIP with stale lineage metadata, rebuild it, and recheck the hosted head after packaging
- keep locally known stale drafts blocked before network access and keep the final hosted compare-and-set authoritative
- reduce the hosted submission success-sound volume from `0.25` to `0.15`

## Problem

Participants often continue working in the same Claude Code or Codex session after sharing it once. The local predecessor can lag the hosted receiver's current logical-session head, causing an otherwise valid updated trace to fail with:

`Declared predecessor does not match the receiver's current head.`

## Safety

- preflight URLs must remain on the configured hosted origin
- submitted shares cannot have predecessor claims rewritten
- exact local content revisions are checked atomically before synchronization
- drafts older than the receiver head remain blocked and require fresh review
- the hosted upload transaction still rejects races that occur after preflight

## Server dependency

Requires the optional preflight capability added by rayward-internal/clawjournal-share#48. Deploy the server PR first. Against older servers, the client retains the existing behavior.

## Validation

- complete Share API class plus revision-tracking tests — 65 passed
- final lineage/preflight regression selection — 4 passed
- frontend test suite — 83 passed
- production frontend build — passed
- `git diff --check` — passed

### DIFF
--- a/clawjournal/parsing/parser.py
+++ b/clawjournal/parsing/parser.py
@@ -3041,7 +3041,8 @@ def _handle_codex_session_meta(
     payload = entry.get("payload", {})
     session_cwd = payload.get("cwd")
     if isinstance(session_cwd, str) and session_cwd.strip():
-        state.raw_cwd = session_cwd
+        if state.raw_cwd == UNKNOWN_CODEX_CWD:
+            state.raw_cwd = session_cwd
         if state.metadata["cwd"] is None:
             state.metadata["cwd"] = anonymizer.path(session_cwd)
     if state.metadata["session_id"] == filepath.stem:
@@ -3067,7 +3068,8 @@ def _handle_codex_turn_context(
     payload = entry.get("payload", {})
     session_cwd = payload.get("cwd")
     if isinstance(session_cwd, str) and session_cwd.strip():
-        state.raw_cwd = session_cwd
+        if state.raw_cwd == UNKNOWN_CODEX_CWD:
+            state.raw_cwd = session_cwd
         if state.metadata["cwd"] is None:
             state.metadata["cwd"] = anonymizer.path(session_cwd)
     if state.metadata["model"] is None:
--- a/clawjournal/web/frontend/src/views/Share/successChime.test.ts
+++ b/clawjournal/web/frontend/src/views/Share/successChime.test.ts
@@ -64,7 +64,7 @@ describe('success sound', () => {
     expect(sound.loop).toBe(false);
     expect(sound.volume).toBe(0.15);
 
-    vi.advanceTimersByTime(900);
+    vi.advanceTimersByTime(1500);
     expect(sound.pause).not.toHaveBeenCalled();
 
     resolvePrimingPlayback();
@@ -72,11 +72,19 @@ describe('success sound', () => {
     await Promise.resolve();
 
     expect(sound.play).toHaveBeenCalledOnce();
-    vi.advanceTimersByTime(899);
+    vi.advanceTimersByTime(1499);
     expect(sound.pause).not.toHaveBeenCalled();
     vi.advanceTimersByTime(1);
+    expect(sound.pause).not.toHaveBeenCalled();
+    expect(sound.volume).toBe(0.15);
+    vi.advanceTimersByTime(250);
+    expect(sound.pause).not.toHaveBeenCalled();
+    expect(sound.volume).toBeGreaterThan(0);
+    expect(sound.volume).toBeLessThan(0.15);
+    vi.advanceTimersByTime(250);
     expect(sound.pause).toHaveBeenCalledOnce();
     expect(sound.currentTime).toBe(0);
+    expect(sound.volume).toBe(0.15);
   });
 
   it('stops silent playback when submission does not succeed', async () => {
@@ -135,7 +143,7 @@ describe('success sound', () => {
 
     expect(sound.play).toHaveBeenCalledOnce();
     expect(sound.pause).toHaveBeenCalledOnce();
-    vi.advanceTimersByTime(900);
+    vi.advanceTimersByTime(1500);
     expect(sound.pause).toHaveBeenCalledOnce();
   });
 });
--- a/clawjournal/web/frontend/src/views/Share/successChime.test.ts
+++ b/clawjournal/web/frontend/src/views/Share/successChime.test.ts
@@ -62,7 +62,7 @@ describe('success sound', () => {
     expect(sound.play).toHaveBeenCalledOnce();
     expect(sound.currentTime).toBe(0);
     expect(sound.loop).toBe(false);
-    expect(sound.volume).toBe(0.25);
+    expect(sound.volume).toBe(0.15);
 
     vi.advanceTimersByTime(900);
     expect(sound.pause).not.toHaveBeenCalled();
@@ -99,7 +99,7 @@ describe('success sound', () => {
     expect(sound.pause).toHaveBeenCalledOnce();
     expect(sound.currentTime).toBe(0);
     expect(sound.loop).toBe(false);
-    expect(sound.volume).toBe(0.25);
+    expect(sound.volume).toBe(0.15);
   });
 
   it('does not restart playback after cancellation wins a priming race', async () => {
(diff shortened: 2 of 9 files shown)


kai-rayward (commented):
Reviewed the lineage reconciliation carefully. The core design holds up: server-authoritative head, the local write guarded by exact create-time content revisions, and the hosted compare-and-set still final. I probed the origin pinning (absolute, protocol-relative, and http-downgrade variants are all rejected), confirmed `already_shared_revision` keeps precedence so `accepted_predecessors` can't suppress a duplicate-upload block, and checked that a server without the capability is a clean no-op — claims aren't even built. The Codex cwd fix is also correct and well-targeted: `_extract_codex_cwd` (which builds the project-index key, i.e. `target_cwd`) takes the *first* cwd while `raw_cwd` was taking the *last*, so a drive-letter case change between `session_meta` and `turn_context` made `state.raw_cwd != target_cwd` at `parser.py:3019` drop the whole session.

I found two issues and pushed fixes for both in 8fcef3b — please review that commit, and revert it if you'd rather take a different approach.

## 1. The preflight egressed session identities before the source-scope gate

`submit_share_to_hosted` checked release/hold state and predecessors up front, but `source_scope_blockers` lives inside `_prepare_share_export_for_upload`, which now runs *after* the first preflight POST. A share holding sessions outside the confirmed source scope put `{session_id, revision_hash, replaces_revision_hash}` on the wire and only then returned 409:

```
status: 409  error: Share contains sessions outside the confirmed source scope
preflight already sent: {'upload_token': ..., 'sessions': [{'session_id': 'sess-0',
  'revision_hash': 'sha256:0f8df299...', 'replaces_revision_hash': None}]}
```

Confirmed source scope is the control that decides whether these sessions may leave the machine at all, so it has to clear before the first request that names them. Pre-PR nothing share-identifying left until after that check. Fixed by checking the scope alongside the other local gates; the export still re-checks it, consistent with how the other gates repeat.

## 2. `expected_lineage` permanently defeated the finalized-export cache

`expected_lineage` is built from every `share["sessions"]` row, but `export_share_to_disk` skips sessions matching `excluded_projects` or with no `get_session_detail` (`index.py:5572-5578`), so the manifest legitimately carries a subset. Set equality was then permanently false:

```
expected_lineage keys: ['trace-drop', 'trace-keep']
manifest_lineage keys: ['trace-keep']
```

Every seal / download / submit for such a share missed the cache and rebuilt — re-running the tiered gate and, with AI-PII review on, a fresh nondeterministic pass. That undoes #161 for those shares and means the bytes reviewed in Package aren't the bytes submitted. Fixed by comparing only the sessions the bundle actually ships: a rebased predecessor still invalidates the cache, and composition changes arrive through `excluded_projects`, which `redaction_settings_fingerprint` already covers.

## Tests

Three added, each verified red before the fix and green after:

- `test_lineage_preflight_waits_for_the_source_scope_gate` — asserts zero preflight requests when the share is out of scope.
- `test_finalized_export_reuse_survives_a_skipped_session` — monkeypatches `export_share_to_disk` to fail on rebuild.
- `test_finalized_export_reuse_rejects_a_rebased_predecessor` — pins the invalidation direction so the subset check can't be loosened into a no-op.

Full suite green on the branch (3629 passed, 1 skipped) and on a trial merge with `main` including #161 (3630 passed). Frontend: 84 passed.

---

## Still open — worth deciding before the server PR ships

**An advertised-but-unavailable preflight hard-fails all manual submission.** A 404 on the endpoint returns 502 and blocks the upload (verified). Combined with the daemon-cached capability document, publishing `manual_lineage_preflight_url` before the endpoint is live — or any endpoint outage — takes down manual submission entirely. Since the hosted compare-and-set stays authoritative, treating 404 (arguably 5xx too) as "unsupported" and falling through to the existing behavior would be strictly safer than failing closed here, and would remove the deploy-ordering coupling the description calls out.

**The rebase guard compares a local clock to a server clock.** `share_created_at > receiver_accepted_at` is the only thing separating "my continued trace" from "claim a head I don't descend from." Two gaps: the client accepts any string for `current_head_accepted_at`, and `_expiry_timestamp` (`daemon.py:1410`) parses a timezone-less ISO string as **local** time. For a participant west of UTC that shifts the receiver's timestamp into the past — the permissive direction. Worth requiring tz-aware and adding a skew margin, given this one comparison carries the whole safety argument.

## Minor

- `_manual_share_lineage_claims` raising `ValueError` for a purely local data problem surfaces as *"Could not verify the hosted trace revision"* / 502. Reachable through `get_share`'s legacy fallback (`index.py:5467`), which returns rows with no `share_content_revision`.
- Two preflight round-trips on every submission, even when the first says `ready` for everything and the package was reused byte-for-byte.
- `RevisionConflictError` on a session-set mismatch (`index.py:4408`) lists every session as a blocker, including the matching ones.
- The description covers commits 1–2 only. The chime change also raises total audible time from 900 ms to 2000 ms (1500 hold + 500 fade), which "reduce the volume" doesn't convey, and the Codex cwd fix is unrelated to share lineage — worth a line in the summary either way.

kai-rayward (commented):
Pushed the preflight fallback in 20a7c7f, rebased onto your 5ee3741. The rebase conflicted in `_synchronize_manual_share_lineage` and at the test insertion point; both resolved by combining, not choosing — please sanity-check that I preserved your intent.

## Fixed: an absent or broken preflight no longer blocks submission

`_lineage_preflight_is_unavailable` separates two shapes from a real answer:

- **endpoint absent** — 404/405/410/501, i.e. the capability document names an endpoint the deployment doesn't serve yet
- **endpoint broken** — any 5xx, i.e. a live endpoint having an outage

Both degrade to `supported: False` and fall through. Auth failures and semantic rejections still surface unchanged, so a 403 on the upload token blocks exactly as before.

Degrading is safe because it reproduces the pre-preflight behavior exactly: no predecessor is rebound, `accepted_predecessors` stays `None` so the local gate keeps its strict reading, and the hosted compare-and-set still refuses a claim that doesn't match the receiver's head. The participant loses the reconciliation, not the protection. This also removes the deploy-ordering coupling in the description — the server PR and this one can now land in either order.

The same commit fixes a composition bug: when the *recheck* was the call that went unavailable, `accepted_predecessors` was overwritten with `None`, so the egress gate refused a share the same request had just rebound. The first response's value is now kept.

Eight tests, each verified red without the fix on your rebased base: the degrade path parametrized over all six statuses, a 403 case pinning that real answers still block, and the recheck-goes-away case. Full suite 3639 passed, 1 skipped.

## Not fixed — a rebased share is permanently unsubmittable after any failed attempt

This one needs a design decision, so I've left it alone.

The rebase persists to `share_sessions.replaces_revision` before the upload. If the upload then fails for any transient reason — connection reset, rate limit, a 5xx from the submissions endpoint, the participant closing the tab — the next attempt is rejected at the **up-front** `share_predecessor_blockers` call, which has no `accepted_predecessors`, and never reaches the preflight that would re-authorize it:

```
attempt 1 -> 502 Could not reach hosted submission service
persisted replaces_revision after attempt 1: sha256:bbbb...
attempt 2 -> 409 Share revisions are duplicate or based on a stale predecessor
attempt 2 blockers: [{..., 'reason': 'stale_predecessor'}]
attempt 2 reached preflight: False
```

The row now holds a receiver-authoritative value that no purely local check can validate, so it reads as stale forever. The happy path only works because the rebase and the upload land in the same request. Recovery means creating a new share.

Two directions, both bigger than a patch on this commit:

1. **Record that a predecessor was server-accepted** — a `share_sessions` column, so a gated migration and a `SECURITY_SCHEMA_VERSION` bump. The up-front gate then distinguishes "stale" from "rebound to the receiver's head."
2. **Roll the rebind back when the submission doesn't produce a receipt** — smaller, but a crash between the write and the rollback lands in the same state, so it narrows the window rather than closing it.

I'd lean towards (1) since it makes the up-front gate correct rather than merely usually-correct, but it's your call.

## On 5ee3741

Reviewed it while rebasing and it reads correctly to me. Scoping `session_ids` to the sessions that actually ship is the right shape, and `packaged_session_ids.issubset(session_ids)` is the anchor that makes it safe — it catches any divergence between what the preflight scoped and what packaging actually emitted, and the final egress gate then runs on the packaged set, so hold state and source scope are both re-checked over exactly what leaves. Worth noting I haven't given it the same adversarial pass as the original commit.

One small thing: when every session in a share is excluded, the up-front `_manual_share_lineage_claims` call raises `Share contains no session lineage claims.` and surfaces that as the 409. The message you added after packaging — `Share contains no sessions eligible for upload.` — describes that case much better; worth using it in both places.

The clock-skew item from my earlier review is still open.

PR #169, opened 2026-07-29
fix(auto-upload): show live enrollment scan progress

### LINKED ISSUE #165: Auto-upload 'Enabling…' has no progress feedback and looks frozen on large session histories
## Problem

Clicking "Enable automatic upload" on the final confirmation screen leaves the button stuck on "Enabling…" for well over a minute with zero progress indication, indistinguishable from a hang. Confirmed via `clawjournal auto-upload status` during the wait — still `off / ready`, i.e. the call is genuinely still in flight server-side, not failed silently.

## Root cause

`enable()` in `clawjournal/auto_upload.py` runs a **strict scan** before it can enroll:

```
# Acceptance matches the displayed profile, so this call intends to
# enroll: refresh the index now — a refresh that just completed for
# the same index and sources is reused, so one interactive flow pays
# for at most one full re-parse.
scan = _strict_scan_for_enable(
    scope["sources"], progress=scan_progress, on_wait=scan_wait_notice
)
```
(`auto_upload.py:1307-1313`)

An earlier comment (`auto_upload.py:1266-1268`) is explicit about the cost: *"the strict refresh re-parses every enrolled source log (minutes of CPU on a large history)."* With a large session history (in this case 118 projects across claude/codex, confirmed via the workbench project list) this genuinely takes minutes.

`enable()` already accepts `scan_progress` and `scan_wait_notice` callables specifically to report progress back to the caller during this refresh. But the HTTP handler that services the frontend's request never wires them up:

```python
def _handle_auto_upload_enable(self) -> None:
    from ..auto_upload import enable

    body = _read_body(self) or {}
    self._send_auto_upload_result(
        enable(
            agent=str(body.get("agent") or "all"),
            ...
            challenge_only=bool(body.get("challenge_only")),
            prepare_for_manual_share=bool(
                body.get("prepare_for_manual_share")
            ),
        )
    )
```
(`clawjournal/workbench/daemon.py:4616-4648` — no `scan_progress=`/`scan_wait_notice=` kwargs passed)

So the progress-reporting mechanism exists in the backend but is dead code from the web UI's perspective — the frontend has no way to show real progress, and the button is left as an indefinite "Enabling…" for as long as the strict scan takes.

This also matches the daemon log evidence from a live repro: the background scanner logged `Skipping background scan: another process holds the scan lock` while the enable request was in flight, confirming the strict scan was actively running and holding the lock — this wasn't a stuck/dead request, just a slow, silent one.

## Why it matters

This is the same "looks frozen" failure mode as #67 (share submission), but for the auto-upload enable flow, and worse here because the strict scan cost scales directly with total history size — the more sessions a user has (i.e. the more they'd actually want automatic recurring coverage for), the longer they stare at an unchanging "Enabling…" button with no way to tell whether it's working or hung.

## Suggested fix

Wire `scan_progress`/`scan_wait_notice` from `_handle_auto_upload_enable` through to the frontend (e.g. via SSE/polling status endpoint, similar to how other long-running operations in this codebase report progress), and update `AutoUploadControls.tsx`'s "Enabling…" state to show real progress (e.g. "Rescanning codex:… (42/118)") instead of a static label.

## Pointers

- `clawjournal/auto_upload.py:1266-1268` — cost comment ("minutes of CPU on a large history")
- `clawjournal/auto_upload.py:1307-1313` — `_strict_scan_for_enable()` call site with unused-by-caller `progress`/`on_wait` params
- `clawjournal/workbench/daemon.py:4616-4648` — `_handle_auto_upload_enable`, missing `scan_progress`/`scan_wait_notice` wiring
- `clawjournal/web/frontend/src/components/AutoUploadControls.tsx` — "Enabling…" button state
- Related: #67 (same UX pattern for share submission)

JustMeStudio:
## Summary
- expose local recurring-enrollment progress through an authenticated polling endpoint
- show hosted capability checks, scan-lock waiting, current source, and project counts while enable is in flight
- cover both the Auto Upload authorization dialog and the combined manual-share enrollment path
- correlate each request with a unique progress ID so retries and multiple tabs cannot surface stale progress
- preserve the existing hosted protocol and CLI behavior

## Validation
- `python -m pytest tests/workbench/test_daemon.py -q` — 225 passed
- `npm --prefix clawjournal/web/frontend test -- --run` — 87 passed
- `npm --prefix clawjournal/web/frontend run build` — passed
- `git diff --check origin/main...HEAD` — passed

Closes #165

### DIFF
--- a/clawjournal/web/frontend/src/api.test.ts
+++ b/clawjournal/web/frontend/src/api.test.ts
@@ -98,6 +98,75 @@ describe('automatic-upload API normalization', () => {
       },
     ]);
   });
+
+  it('tracks accepted enrollment requests and reads their local progress', async () => {
+    const fetchMock = vi.fn()
+      .mockResolvedValueOnce({
+        ok: true,
+        status: 200,
+        json: async () => ({ mode: 'enabled' }),
+      })
+      .mockResolvedValueOnce({
+        ok: true,
+        status: 200,
+        json: async () => ({
+          progress_id: 'progress-issue165',
+          stage: 'scanning',
+          message: 'Refreshing Codex source logs: 42/118 projects',
+          source: 'codex',
+          current_project: 42,
+          total_projects: 118,
+          updated_at: '2026-07-29T00:00:00Z',
+        }),
+      });
+    vi.stubGlobal('fetch', fetchMock);
+
+    await api.autoUpload.enable({
+      agent: 'auto',
+      accepted_authorization_profile_hash: 'profile-hash-v2',
+      progress_id: 'progress-issue165',
+    });
+    await expect(
+      api.autoUpload.enableProgress('progress-issue165'),
+    ).resolves.toMatchObject({
+      stage: 'scanning',
+      source: 'codex',
+      current_project: 42,
+      total_projects: 118,
+    });
+
+    expect(fetchMock).toHaveBeenNthCalledWith(1, '/api/auto-upload/enable', {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        agent: 'auto',
+        accepted_authorization_profile_hash: 'profile-hash-v2',
+        progress_id: 'progress-issue165',
+      }),
+    });
+    expect(fetchMock).toHaveBeenNthCalledWith(
+      2,
+      '/api/auto-upload/enable-progress/progress-issue165',
+      { headers: {} },
+    );
+  });
+
+  it('does not create progress state for a read-only challenge', async () => {
+    const fetchMock = vi.fn().mockResolvedValue({
+      ok: true,
+      status: 200,
+      json: async () => ({ mode: 'off' }),
+    });
+    vi.stubGlobal('fetch', fetchMock);
+
+    await api.autoUpload.enable({ agent: 'auto', challenge_only: true });
+
+    expect(fetchMock).toHaveBeenCalledWith('/api/auto-upload/enable', {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({ agent: 'auto', challenge_only: true }),
+    });
+  });
 });
 
 describe('share scanner recovery API', () => {
--- a/clawjournal/web/frontend/src/api.ts
+++ b/clawjournal/web/frontend/src/api.ts
@@ -25,6 +25,7 @@ import type {
   WorkbenchConfig,
   AutoUploadAgent,
   AutoUploadCandidateReport,
+  AutoUploadEnableProgress,
   AutoUploadHookDiagnostic,
   AutoUploadStatus,
 } from './types.ts';
@@ -364,6 +365,7 @@ export const api = {
       accepted_authorization_profile_hash?: string;
       challenge_only?: boolean;
       prepare_for_manual_share?: boolean;
+      progress_id?: string;
     }): Promise<AutoUploadStatus> {
       const status = await request<Partial<AutoUploadStatus>>('/auto-upload/enable', {
         method: 'POST',
@@ -371,6 +373,12 @@ export const api = {
         body: JSON.stringify(body),
       });
       return normalizeAutoUploadStatus(status);
+    },
+
+    enableProgress(progressId: string): Promise<AutoUploadEnableProgress> {
+      return request(
+        `/auto-upload/enable-progress/${encodeURIComponent(progressId)}`,
+      );
     },
 
     async run(): Promise<AutoUploadStatus> {
--- a/clawjournal/web/frontend/src/autoUploadEnableProgress.ts
+++ b/clawjournal/web/frontend/src/autoUploadEnableProgress.ts
@@ -1,0 +1,43 @@
+import { api } from './api.ts';
+import type { AutoUploadEnableProgress } from './types.ts';
+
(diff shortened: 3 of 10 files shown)


kai-rayward (commented):
Reviewed this closely, mostly around the new endpoint's exposure. It's in good shape — no blocking issues, and one design choice is worth calling out as exactly right: `_handle_auto_upload_enable_progress` touches only an in-memory dict under its own lock and never opens the index. That's what makes polling work at all during a long refresh; had it gone through `open_index()` it would have queued behind the scan's write lock and reported nothing until the work it was reporting on had finished.

Pushed c2dcdba with tests for three behaviors the endpoint already has but nothing held in place. No production code changed.

## Verified

- **Auth.** `_check_api_auth` runs before dispatch for every `/api/*` path, so the new route is bearer-gated by construction, and adding it to the parametrized auth test rather than relying on that is the right instinct.
- **No protocol change.** `enable()` already accepted `scan_progress`/`scan_wait_notice`, so this is wiring rather than a new contract.
- **The refactor doesn't drift.** I diffed the `enable()` argument values across the move into `enable_kwargs` — `profile_hash` is hoisted but computed identically and nothing in the authorization path changed.
- **Progress is confined to real enrollment attempts.** `track_progress` requires a profile hash and not `challenge_only`; confirmed a read-only challenge publishes nothing.
- **The input filter holds.** `../../etc/passwd`, 129 characters, embedded whitespace and non-ASCII all return `400 Invalid progress identifier`, and before `enable()` is called.
- **Memory is bounded** at 100 entries with oldest-first eviction, and the active entry refreshes constantly so it cannot evict itself.
- **Polling lifecycle is right** at both call sites — stop function in `finally`, `active` flag guarding late resolutions, division-by-zero guarded in both consumers.
- Full suite green merged with current `main` (including #168): 3651 passed. Frontend 87 passed, build passes.

## One finding

**The `progress` value in the `enablingAutomaticUploads` branch is never read.** `SubmitStep.tsx` renders the bar from `submitProgress` state, which the timer effect drives off the static `submitStages` array; `currentSubmitStage` is only ever read for `.buttonLabel` and `.detail`. So the new arithmetic —

```js
progress: ... ? Math.min(99, 94 + Math.round(5 * current / total)) : 96
```

— has no effect on what the participant sees. The bar stays on the generic timer, capped at 92 by the `tick`, while the detail text underneath updates with real scan progress.

To be clear about blame: the field is dead on `main` too, so this PR didn't introduce that — it elaborated it. But since driving the bar from real progress is the point of the PR, it's worth either wiring it up or dropping the arithmetic, rather than shipping a computation nothing consumes. Wiring it needs a little care against the 92 cap and the `Math.max` monotonicity in the existing effect, which is why I left it to you rather than guessing at the interaction. Happy to do it if you'd rather.

(I originally read this as a bar that slides backwards — the band starts at 94 against a 96 default — and was about to push a fix. Checking what actually renders showed the value is unused, so there's no visible regression. Flagging the arithmetic itself instead.)

## Minor

- Completed and failed entries are never released, only evicted at the 100 cap. Bounded, so not a leak, but a terminal entry could be dropped once it has been read.
- The poller's `catch` swallows 401 and 500 the same as the expected first-poll 404 race, so a persistent auth failure polls silently every 300ms until the enable settles. The comment explains the 404 case specifically; narrowing the catch to that would make a real failure visible.

## Not a finding, for the record

When `enable()` raises, the connection closes with no HTTP response at all. That is pre-existing — `main` has no handler there either — and this PR's `except` block improves matters by recording a terminal `failed` stage before re-raising.

kai-rayward (commented):
Wired the bar up in 9e56082, since it turned out to need the interaction untangled rather than just a line moved.

## What was in the way

Two things had to hold before the enrollment band could own the bar.

**The crawl would have clawed it back.** `setSubmitProgress((prev) => Math.min(92, prev + ...))` returns `92` for any `prev` above it, so the next 700ms tick after the band reported 97 would have dragged it to 92 and held it there. It now only ever moves forward:

```js
prev >= 92 ? prev : Math.min(92, prev + (prev < 64 ? 3 : 1))
```

**The band started below where the phase already was.** `94 + round(5 * ratio)` opens at 94 against a pre-scan 96, so the first scan update would have moved the bar backwards — a bar losing ground reads as something going wrong. It now runs `96 + round(3 * ratio)`, so it opens exactly where the phase already displayed and sweeps to 99:

```
pre-scan   96
  0/118 -> 96      42/118 -> 97      118/118 -> 99
```

`Math.max` keeps it monotonic against the crawl regardless.

With those in place, `currentSubmitStage.progress` is now a single computed `enrollmentProgress` value that both the label and `submitProgress` read, so the arithmetic can't drift out of use again.

## Making it testable

The bar had no accessible semantics at all — no role, no value — which is both why nothing could assert on it and why it was invisible to assistive tech. It is now a labelled `progressbar` carrying `aria-valuenow`, and your existing test asserts the value reaches 97 at 42 of 118 projects.

That assertion is worth keeping because of what it reads without the wiring: **28**. During enrollment the bar was sitting on `submitStages[1].progress` while the text above it counted real projects.

## Validation

- frontend 87 passed, `tsc --noEmit` clean, build passes
- full Python suite 3653 passed, 1 skipped
- verified the new assertion fails (28 vs 97) with the wiring effect removed, so it guards the actual behavior rather than the arithmetic

Both minor items from my last comment are still open and still just nits: terminal progress entries are only released at the 100 cap, and the poller's `catch` swallows 401/500 identically to the expected first-poll 404.

Issue #198, opened 2026-08-14
add a bug report button inside CL (maybe position at the lower right corner)


richardopenrefinery:
anytime sth goes wrong, users click it, and then it will
1. submit screenshot of current screen of CL.
2. system, software version etc.

JustMeStudio:
Thanks for filing this. I am going to implement it in privacy-bounded stages rather than making a click silently capture and publish the current screen.

ClawJournal's normal session views can contain unredacted transcript text, project/session identifiers, email addresses, local paths, and secrets that deterministic scanning may not have recognized. GitHub issues are public, GitHub's issue API does not provide a safe image-attachment field, and ClawJournal does not currently have a purpose-isolated private support-report receiver. Therefore an automatic screenshot/upload cannot safely be the first increment.

## Phase 1: local, reviewable bug-report draft

The first PR will add:

- A **Report a problem** entry that remains reachable from the normal Workbench, index-recovery/unavailable UI, and React render-error fallback.
- A DB-free, bearer-authenticated `GET /api/support-context` route located before the index-health gate. It will not open or migrate SQLite and will return only a strict allowlist of coarse diagnostic fields.
- Allowlisted context such as ClawJournal version/revision, expected schema version, daemon reachability/health enum, coarse daemon OS/Python/SQLite information, browser family/major version, viewport, and a normalized route template such as `/session/:id`.
- Explicit exclusion of absolute paths, raw URLs/query strings, session/project/share IDs, transcript text, logs, raw error messages/stacks, hostname, username, email, IP/proxy/environment values, tokens, and unknown response fields.
- A dedicated accessible dialog with user-entered summary, reproduction details, and expected behavior. It will generate a fully editable local Markdown draft and show exactly what it contains.
- **Copy draft**, **Download `.md`**, and **Open blank GitHub issue** actions. The GitHub URL will contain no title/body/query data; the user remains responsible for reviewing and pasting/attaching the final report.
- A daemon-unavailable fallback: drafting, copy/download, and the blank GitHub link will continue to work with a fixed `Diagnostics unavailable` marker.
- No screenshot, no upload, no telemetry, and no persistence to localStorage/SQLite in this phase.

The first PR will include backend allowlist/DB-free tests; route and malicious-field sanitization tests; normal/recovery/render-failure launcher coverage; offline/timeout fallback; draft equality tests; and keyboard/focus/accessibility coverage.

## Phase 2: private support intake

[...]

Issue #197, opened 2026-08-12
When the chat history gets too long (>400 chats, long-horizon data), the local index needs repair


richardopenrefinery:
The cause is an overly long chat history — it gets stuck at the rating step. 
One way to optimize this would be to implement a hook that automatically scores items on a daily basis – this way, there’s no need to wait continuously.

Screenshots attached.

<img width="607" height="330" alt="Image" src="https://github.com/user-attachments/assets/b3338deb-c3f2-4f93-9126-d3ebb00aec42" />
<img width="1626" height="1221" alt="Image" src="https://github.com/user-attachments/assets/210b21b2-e26a-469e-a0d5-bcb25353239c" />
<img width="1854" height="966" alt="Image" src="https://github.com/user-attachments/assets/f534904b-b71f-478d-833a-ea61e14b0957" />

<img width="884" height="153" alt="Image" src="https://github.com/user-attachments/assets/b24e0fce-569d-4c49-9498-8b8f33ed4dc2" />

JustMeStudio:
Thank you for the detailed report and for testing ClawJournal in a school-cluster environment. That context is useful.

After reviewing the screenshots and the current client code, I think we should separate two symptoms:

1. **Index recovery:** the repair screen was triggered because SQLite's foreign-key check found an invalid reference in the local index. This is the confirmed direct cause of the screen, but we do not yet know which table is affected or what originally created the orphaned row.
2. **The ~400-session scoring backlog:** this is a separate throughput/scheduling problem. Scoring updates score fields on existing session rows; it does not normally create or delete foreign-key relationships, so we currently have no evidence that the number of sessions itself damaged the index.

The current daemon already performs background scoring in batches of 20 while it is running and the scoring backend has been confirmed. However, each session still requires a separate AI call, sessions are selected newest-first, and per-session failures are not persisted with retry/backoff state. A group of repeatedly failing or timing-out recent sessions can therefore occupy the same batch and starve older sessions. A daily hook would spread the work over time, but would not fix the database inconsistency or the underlying queue behavior. The more direct scoring improvements would be durable retry/backoff, fair scheduling, a total per-session deadline, and visible backlog/progress.

For the current damaged index, **Back up and rebuild index** is the intended recovery action. It backs up the existing database and sidecar files, rebuilds from the original session logs, restores readable review/safety state, and verifies the rebuilt database before normal work and sharing resume. This should repair the current index, although the problem may recur if an underlying storage or external-write issue remains.

[...]

PR #200, opened 2026-08-15
Guard workbench state on network filesystems

JustMeStudio:
## Summary

Relates to #197, but intentionally does **not** close it: the reporter confirmed an NFS-backed cluster home, while the foreign-key failure's original writer is still unproven and the ~400-session scoring symptom is a separate scheduling problem.

This PR makes the confirmed cluster-storage situation safe and diagnosable:

- classifies Linux mounts from `/proc/self/mountinfo` with longest-mount matching, autofs refresh, conservative stacked-mount handling, and a fixed privacy-safe filesystem allowlist;
- refuses every normal mutable/leased index open, creation, or rebuild on a positively identified network/shared filesystem before ClawJournal creates state files;
- rechecks storage at startup, on API health synchronization, and throughout guided recovery so a runtime remount fails closed;
- adds a migration-only recovery screen that tells users to stop ClawJournal, cold-copy the complete `CLAWJOURNAL_HOME`, retain the original, and restart on persistent local storage;
- upgrades interrupted-recovery markers to v2 using a generated backup identity under the current state root, while safely rebasing validated v1 markers after a complete state copy;
- adds `clawjournal --version` and `clawjournal doctor index [--json]` so cluster users do not need an external `sqlite3` binary;
- keeps diagnostics strictly bounded and path/content-free, uses immutable SQLite reads, and declines to open known network storage or a live sidecar snapshot;
- gives both `clawjournal` and `clawshare` a one-line, traceback-free storage refusal.

## Safety and compatibility

- Unknown filesystem kinds remain non-blocking; this avoids guessing on unsupported platforms.
- Positive network-filesystem detection is currently Linux-specific. Windows/macOS remain `unknown` and are not claimed as protected by this PR.
- The recovery flow never deletes the original during migration guidance. Existing v1 backups remain supported, and v2 backup paths cannot escape the current state root through traversal or symlinks.
- Diagnostics omit paths, mount sources, hostname, username, raw errors, session identifiers/content, and user-defined FUSE subtype strings.
- No database schema migration is introduced.

## Non-goals

- This does not implement the proposed durable/fair AI-scoring queue or claim that scoring caused the foreign-key violation.
- This does not add #198's report submission UI or upload screenshots.

## Verification

- Final storage/recovery/diagnostics/CLI subset: **128 passed, 1 platform skip**
- Additional CLI/events/auto-upload/daemon regression run: **655 passed**
- Broader recovery/config/auto-upload-index subset: **268 passed, 15 skipped**
- Frontend: **15 files / 127 tests passed**
- ESLint, TypeScript, Share Queue state test, production Vite build, `compileall`, and `git diff --check` passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,8 @@ clawjournal skill --install-nudge        # opt in to a stale-lessons SessionStar
 clawjournal skill --uninstall-nudge      # disable that reminder and its context output
 clawjournal share --interactive --weekly --no-score # guided sharing without AI scoring
 clawjournal status                       # check your setup
+clawjournal --version                    # show the package version and checkout revision
+clawjournal doctor index --json          # collect bounded, path-free index diagnostics
 clawjournal selfupdate --check           # see whether a newer version is available
 clawjournal selfupdate --reinstall       # update and rerun the installer in one step
 clawjournal selfupdate --reinstall --with-frontend --with-sharing # also add UI + managed scanners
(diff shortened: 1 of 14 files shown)


JustMeStudio:
Follow-up review found and fixed one recovery TOCTOU in 90c9baa: a mount could change from local to NFS after backup but before the old live index was removed. Recovery mutations now recheck storage at marker, manifest, backup, unlink, and cleanup boundaries. Fresh-backup and existing-backup remount regressions both preserve the live index and leave the last durable marker for cold-copy retry.

Post-fix verification: the focused recovery/index/diagnostics set is **208 passed, 1 platform skip**, and an independent reproduction confirms the live DB remains byte-for-byte unchanged with health switching to `storage_migration_required`.

PR #228, opened 2026-09-18
Increase default Sessions selection from 10 to 50

richardopenrefinery:
Sessions now defaults to 50 traces per page instead of 10, so up to 50 visible traces are selected on initial load. This changes the initial page-size value; users can still choose another page size from the existing dropdown.

Updated the existing refill tests for the 50-row boundary, including checks for the default selection count, hidden lookahead row, manual deselection, and Clear behavior.

Validation:

- Frontend tests: 176 passed (`NODE_OPTIONS=--no-experimental-webstorage npm test -- --reporter=dot`; the flag avoids Node 26's native localStorage conflict with jsdom).
- `npm run lint` passed.
- `npm run build` passed, including the share-queue checks and TypeScript compilation.


richardopenrefinery:
wrt checks failtures: the failures come from expired dates in two existing tests. I reproduced both on unchanged main, confirming they’re unrelated to the Sessions change.
- Screenshot recovery: its mock receipt expired September 15, so recovery rejects the screenshot.
- Report listing: its mock reports date from August 16 and now exceed the 30-day retention period, so the list is empty.
The other Python jobs were automatically cancelled after these failures. Windows and frontend checks passed.

